PR #88, opened 2023-05-29
fix: precipitation issue, rmv log mouseparallax, add rule lint

JaimeTorrealba:
- add console error rule in linter
- rmv log in MouseParallax
- fix precipitation error when area not provided

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -18,6 +18,7 @@ module.exports = {
   ignorePatterns: ['**/*.test.ts', 'packages/**/dist', 'package.json'],
   rules: {
     'arrow-parens': ['error', 'as-needed'],
+    'no-console': 'error',
     'comma-dangle': 'off',
     'space-before-function-paren': 'off',
     'max-len': [1, { code: 120 }],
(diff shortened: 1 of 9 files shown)


playground/src/components/PrecipitationDemo.vue
```diff
@@ -66,12 +66,13 @@ pane.addInput(options, 'areaZ', {
...
-      :count="options.count"
       :area="[options.areaX, options.areaY, options.areaZ]"
+      />
+      <!-- :speed="options.speed"
+      :count="options.count"
       :randomness="options.randomness"
-      :size="options.size"
-    />
+      :size="options.size" -->
+      <!--  -->
```
alvarosabu, line 75: Uncomment this part so its available again since it was only a test

.eslintrc.cjs
```diff
@@ -18,6 +18,7 @@ module.exports = {
   ignorePatterns: ['**/*.test.ts', 'packages/**/dist', 'package.json'],
   rules: {
     'arrow-parens': ['error', 'as-needed'],
+    'no-console': 'warn',
```
alvarosabu, line 21: I would suggest doing this one on a separate PR, since is not a fix, is a chore topic. 

You can keep the mouse-parallax log removal, but try in the future to keep 1 PR by 1 scope.

alvarosabu (changes requested):

PR #116, opened 2023-07-01
feat: backdrop stage abstraction

alvarosabu:


public/models/suzanne.json
JaimeTorrealba, line 1: Why do we add models and textures here in cientos? 

Should not go in assets repo?

docs/public/card-bo-bot/CardBoBot-height.png
JaimeTorrealba, line 1: Same here, is there a reason why these textures are here? 

Should not go in assets?

JaimeTorrealba (changes requested):

JaimeTorrealba:
I check the example in netlify deploy, but I'm not sure if is right.

https://deploy-preview-116--cientos-tresjs.netlify.app/guide/staging/backdrop.html

image

public/models/suzanne.json
alvarosabu: I though for docs would be ok to have them locally for performance, but seems like that's breaking when statically generating the assets.

PR #127, opened 2023-07-10
feat(cientos): added useDracoLoaderDispose to useGLTF

Sea-DH1:


src/core/loaders/useGLTF/index.ts
```diff
@@ -70,3 +70,9 @@ export async function useGLTF(
 ): Promise<GLTFResult> {
   return (await useLoader(GLTFLoader, path, setExtensions(options, extendLoader))) as unknown as GLTFResult
 }
+
+export function useDracoLoaderDispose() {
```
alvarosabu, line 74: Hi, @Sea-DH1 thanks for opening the issue and providing a solution.

What if we save the user some trouble and dispose the dracoLoader using `onMounted` hook inside of this composable? That way the end user doesn't need todo it manually

alvarosabu (changes requested):

src/core/loaders/useGLTF/index.ts
Sea-DH1: The main reason is that a decoder in it needs to be loaded. If it is disposed of, it will be loaded again next time

Sea-DH1: > 
I forgot, it can be called there when the webglrenderer is destroyed, so that the user does not have to call it manually

JaimeTorrealba: @Sea-DH1 Hi, could you finally solve this issue?

Sea-DH1: @JaimeTorrealba 
Hi, I have tried but cannot get the context of TresContext so that I cannot dispose when the WebGLRenderer is destroyed.

JaimeTorrealba: Ok, we have this one available https://docs.tresjs.org/api/composables.html#usetrescontext
But that will only work on the component instance, for the composable. We couldn't have the context yet... In that case, I'll say dispose on the life cycle.

Sea-DH1: If it is processed in the life cycle, it seems that it cannot be automated, right?

JaimeTorrealba: I don't know any method for that, sorry

hawk86104 (commented):
In this way, after obtaining the model, the memory resources of dracoLoader are automatically released

JaimeTorrealba:
@hawk86104 Would you like open another issue fixing this one? 

I can close this issue

JaimeTorrealba:
Done with https://github.com/Tresjs/cientos/pull/383

PR #138, opened 2023-07-19
fix: background not checked right useEnvironment

Magicpig55:
useEnvironment composable isn't checking the background property correctly. Since it's checking if it is not equal to undefined, and it always defaults to false, the scene background will always be set. This PR adjusts it so the background will only set if the property is truthy. 

Example of the bug here: https://stackblitz.com/edit/tresjs-basic-pvijjg?file=src%2Fcomponents%2FTheExperience.vue


alvarosabu:
Hey @Magicpig55 thanks a lot for this!

PR #142, opened 2023-07-22
chore: add reactive method isLock

JaimeTorrealba:


alvarosabu (changes requested):
Nice work. Although Im not entirely sure Im comfortable with the use of `v-model` here. Especially because there is no use of `modelValue` inside of the component, the idea of `v-model` is to provide 2-way binding, here is just a replacement for event emission 

I suggest you define a `lock` and an `unclock` event so the user controls the isLock logic outside.

Regarding the PR title and the commit the use of `chore` is incorrect. This represents an end-user change (the addition of a v-model API or the new events) so therefore is a `feat`

A `chore` is meant for internal changes like refactors or utils that don't affect the end user in terms of documentation or API. This will not trigger a new version which it should

PR #154, opened 2023-07-31
feat: add camera controls to new context

JaimeTorrealba:


src/core/controls/CameraControls.vue
```diff
@@ -360,16 +360,16 @@ const subsetOfTHREE = {
...
 watch(controlsRef, value => {
   addEventListeners()
   if (value && makeDefault.value) {
-    setState('controls', value)
+    controls.value = value
   } else {
-    setState('controls', null)
+    controls.value = null
   }
 })
```
Tinoooo, line 375: I know this didn't change in this PR, but I leave this comment anyways 👹. It would be cool if the emits would be typed here. Something like this:

```
const emit = defineEmits<{
  'change': [controls: CameraControls]
  'start': [controls: CameraControls]
  'end': [controls: CameraControls]
}>()

```

I am also fine if you leave it the way it is. The change shouldn't hold us from merging this PR ☺️.

JaimeTorrealba: I'll skip this for now, 
I have some errors and all the other components are the same, if you agree we can change it latter.

Cientos is not well typed anyway, (we need to change that in some future refactors)

Tinoooo (approved):

Issue #171, opened 2023-08-13
LensFlare


JaimeTorrealba:
### Description

As a developer using Tresjs I would like to use lensFlare easily

### Suggested solution

Create an abstraction of the ThreeJs addons [Lensflare](https://threejs.org/docs/index.html#examples/en/objects/Lensflare)

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/cientos/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/cientos/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://cientos.tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/cientos/issues) that reports the same bug to avoid creating a duplicate.

andretchen0:
@JaimeTorrealba 

I'll take a crack at it.

JaimeTorrealba:
@andretchen0 assigned my friend.

Thanks and any question, we're here to help

andretchen0:
## Visuals

Here's what I've got at the moment, visually. How's that looking?

https://github.com/Tresjs/cientos/assets/20469369/c932eb8b-9564-4e06-8923-4bf89b2696b5


----

## Textures

I'm not sure about the best approach for the default textures.

Users can supply their own textures, but I'd prefer to include some sensible defaults.

There are textures in the Three.js repository. But they're licensed as [CC non-commercial](https://github.com/mrdoob/three.js/blob/67d574bcda95fd50bc3477084eb9d092a3004d79/examples/textures/lensflare/LICENSE.txt#L4).

So, at the moment, I'm generating the textures in a canvas at runtime. It's about 150 lines of TS – over half of the LOC in the component. Is that ok, or would you prefer using image textures?

JaimeTorrealba:
Hi @andretchen0 it looks amazing, I love it

For the texture, I recommend that you use image texture, and set it as a default.

You can look at the [smoke](https://github.com/Tresjs/cientos/blob/main/src/core/abstractions/Smoke.vue) element as a reference. In that component, if you don't provide a specific texture, it uses one as default (that means you always have to use a suspense component, and wrap it, remember to add it to the docs)

[...]

PR #185, opened 2023-08-25
feat: lensflare

andretchen0:
@JaimeTorrealba 

Here's the Lensflare as it stands, for discussion. 

## Playground

If you start the playground, the lensflare demo is at http://localhost:5173/abstractions/lensflare

If you refresh the demo, the large "star" in the middle gets a new seed, so it'll look slightly different.

## Lensflare

`src/code/abstractions/Lensflare.vue` could shed about 200 LOC if we dropped the seedable pseudorandom generation. About 150 LOC is setting up then selecting the seeded pseudorandom defaults.

But those 200 lines make it a lot easier for the user to add a Lensflare.

[As mentioned here](https://github.com/Tresjs/cientos/issues/171), a user can currently do this

```
<Lensflare :seed="1918384" />
```

and it'll produce a Lensflare with about 25 elements based on the seed – always the same for the same seed.

This

```
<Lensflare :seed="1918384" :flare="[{texture:'https://myfakeurl.com/images/texture.png'}]" />
```

will replace the seeded first element's texture with the image at the passed url.

If we remove the pseudorandom generation, the user will have to do something like:

```
 <Lensflare :flare="[{ texture: circleBlurTexture, size: 366, distance: 0, color: 'white' },
  { texture: circleTexture, size: 155, distance: -0.3, color: 'gray' },
  { texture: starTexture, size: 512, distance: 0, color: 'gray' },
  { texture: lineTexture, size: 512, distance: 0, color: 'white' },
  { texture: ringTexture, size: 512, distance: 0, color: 'white' },
  { texture: circleTexture, size: 256, distance: 0.6, color: 'gray' },
  { texture: circleTexture, size: 80, distance: 0.7, color: 'gray' },
  { texture: circleBlurTexture, size: 70, distance: 0.9, color: 'white' },
  { texture: circleTexture, size: 65, distance: 1, color: 'white' },
  { texture: circleBlurTexture, size: 60, distance: 1.8, color: 'white' },
  { texture: circleTexture, size: 60, distance: 2.0, color: 'gray' },
  { texture: circleTexture, size: 155, distance: 2.3, color: 'gray' }]" />
```

### Note

The THREE Lensflare isn't really meant to be updated. It makes its elements private. 

I've made the elements reactive to prop changes by hanging on to their references, but updating them is a rather slow process. User input has to be merged with seeded defaults or a default element, then that has to be merged with the THREE Lensflare's elements.

It may be slow if done every frame. Since that's not the way the THREE implementation appears to be meant to work, I didn't worry about it.

----

## RandUtils

For `src/utils/RandUtils`, I reused the parts of THREE's `MathUtils`, but made them use the seedable pseudorandom generator. If a module uses `RandUtils` and always calls methods the same way, it'll always produce the same numbers, array samples, etc. I *think* that could be pretty handy for a project like Cientos, but I don't want to force it on anyone. FWIW, I got the idea from Apple Motion, which takes random seeds for a lot of its objects.


src/core/abstractions/Lensflare.vue
```diff
@@ -0,0 +1,305 @@
...
+  });
+
+  return [...oversizeElements, ...bodyElements, ...frontElementProps, ...backElements];
+}
+
+const defaultElement: LensflareElement = {
+  texture: `${TEXTURE_PATH}cirlceBlur.png`,
+  size: 64,
+  distance: 0,
+  color: new THREE.Color("white")
```
JaimeTorrealba, line 207: We can't pass the color as a prop?

JaimeTorrealba (changes requested):
First let me tell you thanks for this amazing contributions, I believe almost all the changes are just minimal stuff

Seems really cool, the effect

Even so, in addition to the requested changes:

All the props should be reactive, 
I was trying to changes the seeds from 0 to 10000, but I can't see any visual changes
Have you tried adding your own textures¿? 
We could consider clean a little to the component and extract some logic into other files, what do you think?
The final result could be something like:
abstractions/
     LensFlare/
          index.vue
          Utils (or any other name) // And here put some logic so we 



Sorry for the time to response is a big component, but I'm excited to add this to Cientos 

Also, I'm on the official TresJs Discord for a more quick respond

andretchen0:
> First let me tell you thanks for this amazing contributions, I believe almost all the changes are just minimal stuff
> 
> Seems really cool, the effect
> 
> Even so, in addition to the requested changes:
> 
> All the props should be reactive, I was trying to changes the seeds from 0 to 10000, but I can't see any visual changes Have you tried adding your own textures¿? We could consider clean a little to the component and extract some logic into other files, what do you think? The final result could be something like: abstractions/ LensFlare/ index.vue Utils (or any other name) // And here put some logic so we
> 
> Sorry for the time to response is a big component, but I'm excited to add this to Cientos
> 
> Also, I'm on the official TresJs Discord for a more quick respond

Thanks for the feedback. I'll go through it and make the changes.

src/core/abstractions/Lensflare.vue
```diff
@@ -0,0 +1,305 @@
...
+
+const props = withDefaults(
+  defineProps<{
+    flare?: FlareElementProps[],
+    scale?: number,
+    seed?: number,
+  }>(),
+  {
+    flare: undefined,
+    seed: undefined,
```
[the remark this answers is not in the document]
andretchen0: > seeds is ok to be undefined here?

The way I've currently defined it, an `undefined` seed has a different meaning than `0`.

Below ...

* `hasSeed` means the seed has a number value
* `hasFlare` means the user has passed flare options.

In the following case, it works like you suggest – no seed was provided, but it defaults to 0:

| hasSeed | hasFlare | Effect |
| --------- | --- | --- |
| `false` | `false` | Use seeded elements with seed == 0 |

But here's the exception, where an `undefined` seed is meaningful:

| hasSeed | hasFlare | Effect |
| --------- | --- | --- |
| `false` | `true` | Use user flare, fill missing values with defaults |

This is the use case I had in mind: 

The user wants to try to create a particular lensflare. 

They don't pass a seed, because they want to create something specific, not random. But they don't want to have to provide all of the options for each `FlareElementProps`, if they like the single `defaultElement`.

I thought defaulting to a single `defaultElement` in this case would make things more predictable for the user.

----

I should say that I'm not attached to things being this way. It's just what made sense to me while writing the code.

I'll follow your lead here.

src/core/abstractions/Lensflare.vue
```diff
@@ -0,0 +1,305 @@
...
+  } else {
+    return n1 * (x -= 2.625 / d1) * x + 0.984375;
+  }
+}
+
+const easingFunctions = [linear, easeInCubic, easeInOutCubic, easeInQuart, easeOutBounce];
+
+const lerp = THREE.MathUtils.lerp;
+
+const TEXTURE_PATH = 'https://raw.githubusercontent.com/andretchen0/tresjs_assets/' +
```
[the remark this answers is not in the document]
andretchen0: Sure thing. I might need to change some images so I'll keep this in my fork for now. But once it's ready for review, I'll submit a PR to the assets repo.

src/core/abstractions/Lensflare.vue
andretchen0: Are we talking about a 'global' color prop set on the whole flare to color all the elements? Like so?

```
<TresPointLight :color="new Color(1, 1, 1)" :position="[10, 5, 0]">
  <Lensflare :color="new Color(0, 0, 1)" />
</TresPointLight>
```

## If so, please confirm:

I assume that if a 'global' color prop is set by the user, it ...

* would override the colors of the seeded random values 
* would be overridden by the user provided `flare` array options, e.g.,:

```
<TresPointLight :color="new Color(1, 1, 1)" :position="[10, 5, 0]">
  <Lensflare :color="new Color('blue')" :flare="[{color:'white'}, {color:'yellow', size:64}, {size:124}]" />
</TresPointLight>
```

would have three flare elements: a white, a yellow, and a blue.

andretchen0:
@JaimeTorrealba 

I've made most of the requested changes above. Let me know about the remaining issues and we'll move forward.

You can check out the playground at:

http://localhost:5173/abstractions/lensflare

Fwiw, I had earlier included `RandUtils` in `src/utils`, but I don't want to be presumptive about its use. I've moved it into the new `src/core/abstractions/Lensflare` folder.

src/core/abstractions/Lensflare/easing.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (x < 1 / d1) {
+    return n1 * x * x;
+  } else if (x < 2 / d1) {
+    return n1 * (x -= 1.5 / d1) * x + 0.75;
+  } else if (x < 2.5 / d1) {
+    return n1 * (x -= 2.25 / d1) * x + 0.9375;
+  } else {
+    return n1 * (x -= 2.625 / d1) * x + 0.984375;
+  }
+}
```
JaimeTorrealba, line 30: I found this really useful, but as you say in other conversation

I would move this to utils, other abstraction, could use them

src/core/abstractions/Lensflare/RandUtils.ts
```diff
@@ -0,0 +1,133 @@
...
+    return () => {
+      // NOTE: Mulberry32 generator
+      seed += 0x6d2b79f5
+      let t = seed
+      t = Math.imul(t ^ (t >>> 15), t | 1)
+      t ^= t + Math.imul(t ^ (t >>> 7), t | 61)
+      return ((t ^ (t >>> 14)) >>> 0) / 4294967296
+    }
+  }
+}
```
JaimeTorrealba, line 133: I would let RandUtils inside the Lensflare folder

src/core/abstractions/Lensflare/index.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  // NOTE:
+  // Flare elements are divided into back, oversize, body, front.
+  // They are arranged as such, relative to the light source and camera:
+  //
+  // [distance < 0]   [distance == 0]   [distance > 0]
+  //                      light                          camera
+  //      back        body, oversize       front
+
+  const oversizeTexturesOptional = [line, ring]
```
JaimeTorrealba, line 105: We could give the possibility to the final user to tweak this values, adding them as a prop, in an object format
for example:
const oversize = {
    colors: ["white"]
    ...
}

And so on, and then provide props, so the use could achieve and understand this

I would like to reduce as much as possible hardcode params. Also, if you feel comfortable you can create a constant.ts inside the folder.

JaimeTorrealba (changes requested):
I like the re-organization of the files, we're close to finish this :)

src/core/abstractions/Lensflare.vue
JaimeTorrealba: You're right could be a mess, let's keep what you have here

src/core/abstractions/Lensflare.vue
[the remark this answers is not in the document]
JaimeTorrealba: In that case, we should provide good docs which explain this case, I found a little hard if the user doesn't know how lensflare works

src/core/abstractions/Lensflare/index.ts
andretchen0: > I would like to reduce as much as possible hardcode params.

Sounds good.

I think it'd be really powerful to hand the random presets over to the user. In between other things, I've been poking around with it for a few days, but I find the TS typing is a bit tricky and ends up being unintuitive.

For now, I'll follow your suggestion to put the values in a `constant.ts` file and push. If another module requires the random seeded tools, maybe it can be revisited at that point.

andretchen0:
I think I've covered what was discussed. Please let me know if I've misunderstood or forgotten anything.

----

@JaimeTorrealba , you said ...

> We could give the possibility to the final user to tweak this values, adding them as a prop, in an object format
for example:

````
const oversize = {
colors: ["white"]
...
}
````

That's possible now. The seeded props is just an array of entries like this:

````
export const oversize: SeedProps = {
  texture: [line, ring],
  color: ['white'],
  distance: [0, 0],
  size: [750, 1024],
  length: [0, 2],
}
````

There's no longer different logic for generating e.g., "oversize" vs. "front" elements. There's just the different entries in `SeedProps[]`. Hopefully, that makes things cleaner and easier to understand.

----

I incorporated a Leches GUI in the demo with many of the random seeded props – though not all as there are too many of them to fit on my screen.

**Comment by alvarosabu** [comment]
> I incorporated a Leches GUI in the demo with many of the random seeded props – though not all as there are too many of them to fit on my screen.

@andretchen0 You can create collapsable folders, I forgot to add it to the docs

```ts
const { value } = useControls('folderName', { 
   control1: { value: 1, min:0, max:2, step:0.1 }
})
 ```

**Comment by alvarosabu** [comment]
@JaimeTorrealba Since I added @andretchen0 to the repo, maybe it will be easy if he creates a branch `feat/171-lensflare` on this repo instead of the fork and just copy all his changes there so we can download it and test it locally

wdyt?

**Comment by JaimeTorrealba** [comment]
@alvarosabu haha, knew you'd enjoy seeing Leches being used :)

Regarding stopping using the fork or creating a new branch directly in the repo, I would say it depends 100% on @andretchen0, It will definitely be easier to test, but we're almost finished here. We just need the docs and the asset links as far I can tell

File: src/core/abstractions/Lensflare.vue (hunk 1 above)
Comment (andretchen0) [reply src/core/abstractions/Lensflare.vue]: Just FYI, I think this is a good idea, so I've added it.

In addition to the previous props, the user can now also specify any of the props on here:

```
<Lensflare color="blue" :distance="1" :size="512" :texture="[... some path]" />
```

I'll push tomorrow, so you can have a look.

**Comment by andretchen0** [comment]
@JaimeTorrealba

## New props: color, distance, size, texture

As mentioned, the user can now put these props directly on the component: color, distance, size, texture.

They serve as defaults. 

* Properties present in user-specified `:elements` will overwrite the defaults.
* The defaults overwrite properties on seeded random elements.

It's handy in a case like this:

```
// Generate some seeded random element and make them all red.
<Lensflare color="red" :seed="99999" />
```

Or:

```
// User sets defaults and then only specifies what differs, using :elements
<Lensflare color="gray" :distance="0" :size="128" :texture="circle" :elements="[
    {color: "white", size: 512},
    {color: "white", texture: rays, size: 256},
    {distance: 0.1},
    {distance: 0.14},
    {distance: 0.16},
    {distance: 0.175},
]" />
```

### Confusion? 

#### Please let me know what you think here. 👈

One potentially confusing case is this:

```
<Lensflare color="red" :distance="0" :size="256" texture="http://..." :seed="99999" />
```

The user has specified a seed, which would normally trigger seeded random element generation. BUT, I ignore it. That's because by specifying all these `{color, distance, size, texture}`, all fields on the seeded random elements would be overwritten.

So, I chose to ignore the seed and just show a **single element** specified by  `{color, distance, size, texture}`.

### Less confusing alternative?

The alternative would be to use the ~20 randomly generated elements, but overwrite all the fields. That leads to a stack of ~20 of the same elements on screen.

----

Feedback is appreciated.

Let me know if this is too confusing.

andretchen0:
@JaimeTorrealba 

I've added the docs at http://localhost:5173/guide/abstractions/lensflare.html

Still waiting on https://github.com/Tresjs/assets/pull/7 but once that's done and the texture paths are updated here, this is ready to go, unless there are more requested changes.

JaimeTorrealba:
@andretchen0 perfect, I already approved the https://github.com/Tresjs/assets/pull/7
So I'll make some last test and we're ready to go :)

andretchen0:
I've updated the paths to point at Tres Assets. 

The Tresjs/assets readme says:

> Repository containting static assets for TresJS demos on [Stackblitz](https://stackblitz.com/@alvarosabu/collections/tresjs)

I use those assets as defaults for the Lensflare. That means they will likely be used in Cientos' users' software. Is that ok, or should they be placed somewhere else?

JaimeTorrealba:
I have to say, incredible work!!! It's a huge contribution

When you feel ready, please remove the draft-status and let's merge it 

Also, if you want, the demo component for lensFlare will be really useful in the [Playground](https://github.com/Tresjs/playground) let me know if you would like to add it there (after finalizing this PR of course)

andretchen0:
> BTW all the images you have used are free of charge right?

Yes. I made them using GIMP and the canvas API.

They are free to use and unencumbered by licenses.

JaimeTorrealba (approved):

andretchen0:
> Also, if you want, the demo component for lensFlare will be really useful in the [Playground](https://github.com/Tresjs/playground) let me know if you would like to add it there (after finalizing this PR of course)

Sure, that would be great.

Would you like me to fork the playground repo?

JaimeTorrealba:
Yes, but you're now part of the Tres org right?
Now you don't need to fork, just download the repo :)

andretchen0:
Oh, sorry. I had to fork tres-assets because I wasn't part of that team.

Ok. I'll just create a branch then. Thanks!

PR #187, opened 2023-08-28
feat: SVG

andretchen0:
@JaimeTorrealba 

Here's a draft of the SVG feature. Start the playground then see `http://localhost:5173/loaders/SVG`.

[There appears to be a conditional rendering bug in Tres.](https://github.com/Tresjs/tres/issues/303) Putting conditional logic in the template and then changing props triggers the bug.

I dealt with that by simply recreating geometries, materials, etc. in the setup script whenever a prop changes. Ideally though, this would happen on the component side; that way the system can simply patch what's updated, rather than recreate the whole component.

Feedback/advice is welcome!


playground/src/pages/SVGDemo.vue
```diff
@@ -0,0 +1,66 @@
...
+    <TresGridHelper :args="[10, 10]" />
+    <TresGroup :scale="0.01" :position="[-2.1, 1, 0]">
+      <Suspense>
+        <SVG :src="svgTriangleString" :skip-fills="skipFillsA" />
+      </Suspense>
+      <Suspense>
+        <SVG :src="svgSquareString" :skip-fills="skipFillsB" />
+      </Suspense>
+      <Suspense>
+        <SVG :src="svgHeartURL" :skip-fills="skipFillsC" :position="[321.5, -4, 0]" />
```
JaimeTorrealba, line 59: 321.5 in position 😮  what is that jajaja

src/core/loaders/SVG/component.vue
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  layers.value = _layers;
+}
+
+</script>
+
+<template>
+  <TresGroup v-bind="$attrs" ref="groupRef">
```
JaimeTorrealba, line 153: Here we don't need the $attrs as is just one component will inherit by default

And let's try to add another name groupRef something more meaningful, remember that we're going to expose this name

src/core/loaders/SVG/component.vue
```diff
@@ -0,0 +1,159 @@
...
+const groupRef = shallowRef();
+const layers = shallowRef([] as SVGLayer[]);
+const paths = shallowRef([] as SVGResultPaths[]);
+
+defineExpose({ value: groupRef });
+
+watchEffect(async() => useSVG(src.value).then(SVGResult => paths.value = SVGResult.paths));
+watch([skipFills, skipStrokes, fillMaterial, strokeMaterial, paths], updateLayers);
+
+async function useSVG(src: string) {
```
JaimeTorrealba, line 96: Let's add some warning in case we can't load the resource (I believe here is a good option but is up to you)

You can use the useLogger() and we have the standard for the logs

JaimeTorrealba (changes requested):
I love the demo 😄 

Thanks for your contribution, this will be added to the next release possible

playground/src/pages/SVGDemo.vue
andretchen0: The triangle and square SVGs were pulled from the three-part logo, so they have the proper offsets when displayed at `[0, 0, 0]`. 

The Cientos heart SVG has its local `0, 0` at the top left of the heart, so it has to have a `:position` to be placed in the right spot. I thought this was a good idea though, as it demonstrates that the SVG can take a `:position`. 

I can bake the `:position` into the heart SVG coordinates and then remove the `:position` in the demo, if that's what you'd prefer. Let me know.

src/core/loaders/SVG/component.vue
andretchen0: Good catch. Thanks. Will rename `groupRef` -> `svgRef`.

andretchen0:
@JaimeTorrealba 

I was playing around with the Drei `Svg`, which I based the Cientos `SVG` on. It's got some weirdness in the proper conditions:
Screenshot 2023-08-30 at 03 02 22

Try orbiting around in this scene:

https://codesandbox.io/s/frosty-panka-jm5r99?file=/src/App.js

----

Three

The [Three example](https://threejs.org/examples/#webgl_loader_svg) gets around that by setting the render order for each mesh/layer:

https://github.com/mrdoob/three.js/blob/b447ff2911cbefcd650df5a18a905e7df882fc78/examples/webgl_loader_svg.html#L193

But changing render order might not be what users expect either. Higher render ordered elements always render later – "on top" of earlier rendered elements.

----

Another option:

Both Three and Drei set the material depthWrite to false. We could set it to true (the default). That will lead to z-fighting, because by default, all layers are rendered at z = 0. But we *could* stack the layers on the z, incrementing by some small amount for each layer.

The downside of that is that orbiting behind the SVG will show the bottom of the stack (which is likely to be ugly).  Here's the tiger in our demo using this technique.

Front:

Screenshot 2023-08-30 at 02 59 40


Back:

Screenshot 2023-08-30 at 02 59 46

----

No perfect options, it seems.

Maybe we should just stick with the way Drei approached things.

What do you think?

andretchen0:
* Renamed the exposed ref -> `svgRef`
* Dispose geometries when props change or when component unmounts
* Removed unneeded `key` in `layers`

andretchen0:
> @andretchen0 what if we let the final user choose? 
> If not, let stick with the drei approach

I've added a `depth` prop. The possible values are:

* `flat` (default) : this is the Drei approach
* `renderOrder`: this is the THREE approach
* `offsetZ`: this increments the geometry on the z by a small value for each 'layer'
* number: same as `offsetZ`, but allows the user to specify the offset

There are tiger heads in the demo showing each option. Orbit around to see the advantages/disadvantages of each.

At least for me, the `offsetZ` version z-fights a little bit. When the scale is 1 it doesn't z-fight. I left the scale and z-fighting to highlight the problem. That's why I added the possibility to pass a `number` to set the offset.

The `number` option is exaggerated for effect in the demo – it works the same as `offsetZ`.

JaimeTorrealba:
@andretchen0 that's just perfect!!!
Great work 

Please send the docs and remove the draft state

andretchen0:
Hey @JaimeTorrealba 

I'm keeping this as a draft for the moment, as I'd like your feedback.

## Docs

**How do the docs look?**

http://localhost:5173/guide/loaders/svg.html

I spent a lot of time explaining the `depth` prop and added some troubleshooting. I don't know if that's necessary or not.

## Depth disadvantages demo

I've also added a `depth` demo to the playground, showing the disadvantages of each setting:

http://localhost:5173/loaders/SVG/depth

**Is this useful? If so, should it be added to the docs somehow?**

## Depth

**I propose we change the default `depth` prop value to `'renderOrder'`. Is that ok with you?**

That's how the example on the THREE website is configured. It's the most dependable for actually rendering individual SVGs. 

It changes the render order though, so if there are multiple overlapping SVGs with `depth="renderOrder"`, they'll appear to be interleaved. For example:

Screenshot 2023-08-31 at 16 10 33

JaimeTorrealba:
@andretchen0 Hi.

Sorry for the slow response, I've been at a conference.

1.- The docs and the explanation look good (I have an issue adding inline examples, so don't worry about it)
2.- Thanks for this, I'll merge them and check them out, but in the future we just need one demo per feature
3.- It sounds better. Change the depth default to renderOrder (remember to update the docs)

*As you use a fork I can't easily check out to your branch, that's why it takes time

I'll proofread the docs. If you make the necessary changes for the depth default, we are ready to go :)

docs/guide/loaders/svg.md
```diff
@@ -0,0 +1,88 @@
...
+| **skipStrokes**     | `boolean`                                        | If `true`, the SVG strokes will not be rendered.                          | `false`     |
+| **skipFills**       | `boolean`                                        | If `true`, the SVG fills will not be rendered.                            | `false`     |
+| **strokeMaterial**  | `MeshBasicMaterialParameters`                    | Props to assign to the stroke materials of the resulting meshes.          | `undefined` |
+| **fillMaterial**    | `MeshBasicMaterialParameters`                    | Props to assign to the fill materials of the resulting meshes.            | `undefined` |
+| **strokeMeshProps** | `TresOptions`                                    | Props to assign to the resulting stroke meshes.                           | `undefined` |
+| **fillMeshProps**   | `TresOptions`                                    | Props to assign to the resulting fill meshes.                             | `undefined` |
+| **depth**           | `'flat' \| 'renderOrder' \| 'offsetZ' \| number` | How should the meshes be "stacked"? ([See "Depth"](#depth))               | `flat`      |
```
JaimeTorrealba, line 32: So as we agree, remember to make the necessary changes here.

Stacked is fine, should not have the question mark

JaimeTorrealba (changes requested):

docs/guide/loaders/svg.md
andretchen0: Thanks for the feedback. Will make the changes soon and push.

> *As you use a fork I can't easily check out to your branch, that's why it takes time

I'd like to make things easier for you. Is there another workflow you prefer?

JaimeTorrealba: If you like, and you plan to do more PRs to Tresjs ecosystem, I could add you to the organization (with no obligation or responsibility)?

andretchen0: > you plan to do more PRs to Tresjs ecosystem

I was going to propose working on `<Sky />` after these PRs are merged.

Sure, go ahead and add me to the organization if that works for you.

JaimeTorrealba (approved):
Great job! 👍

alvarosabu:
Hi @andretchen0, Jaime and I where wondering if you would like to join the org as contributor since you are adding a lot of value to the ecosystem with this PRs, it will make it easy also because you would not need to fork the project.

Wdyt?

By the way, are you in the discord channel?

Cheers

andretchen0:
> Wdyt?

Sure thing.

> By the way, are you in the discord channel?

Not yet. I'll check it out.

PR #201, opened 2023-09-06
feat: add distance prop in camera controls

notarun:
CameraControls abstraction was missing distance prop, this PR adds it. I have used the closed issue number here, should i create a new issue for that?


JaimeTorrealba (approved):
Everything fine :)

Thanks for this

PR #203, opened 2023-09-06
feat: add stats-gl component

### LINKED ISSUE #197: Stats-gl
### Description

Add a component, stats-gl, to measure the performance of the scene

### Suggested solution

See https://github.com/pmndrs/drei#statsgl

### Alternative

_No response_

### Additional context

_No response_

### Validations

notarun:
Closes #197 
Still need to add docs I guess 😅


JaimeTorrealba:
Seems everything fine :)

I'll wait for the docs

JaimeTorrealba:
@alvarosabu Can you help me here? Would be a good idea to move them to peer dependencies?

alvarosabu:
@notarun @JaimeTorrealba what is the difference between stats.js and stats-gl?

JaimeTorrealba:
@alvarosabu stats-gl is a new fancy monitor, with more capabilities than stats.js

https://github.com/RenaudRohlinger/stats-gl

It was adopted by R3F and people like it

alvarosabu:
That being said, make sense to have both @JaimeTorrealba ?

JaimeTorrealba:
I would say yes @alvarosabu, more option to the user is always better

We could think better in the final bundle-size if that's the problem, move to peer-deps (but with clear communication to the users, to avoid the useTweakpane issue)

JaimeTorrealba (approved):
Excellent work

PR #207, opened 2023-09-10
fix(cientos): controls' props priority

damiankoper:
It appears that the `domElement` and the `camera` props cannot be overwritten in `Controls` components. This PR fixes it.


JaimeTorrealba (approved):
I see everything fine.
Thanks for this PR :)

PR #212, opened 2023-09-13
fix(Levioso): changing speed causes jerkiness

### LINKED ISSUE #208: Levioso – changing speed leads to (likely) unwanted behavior
### Describe the bug

If Levioso's `speed` prop is updated, it leads to (likely) unwanted behavior.

Here it's changed over time:

https://github.com/Tresjs/cientos/assets/20469369/876fdff9-6e45-43b2-a6ef-2440cb8d3508

The problem is [here](https://github.com/Tresjs/cientos/blob/cc3f3e98a4d0bb9fd22a2120fa529a7db081b5cf/src/core/abstractions/Levioso.vue#L34C1-L40C60):

````
onLoop(({ elapsed }) => {
  if (!groupRef.value) return

  const theta = offset.value + elapsed
  groupRef.value.rotation.x = (Math.cos((theta / 4) * props.speed) / 8) * props.rotationFactor
  [ ... ]
````

Doing a quick substitution and dropping the divisors and `offset`, we have:

```
theta = elapsed * props.speed
```

So `props.speed` scales `elapsed` time.

If instead, we scale `delta` time, we get the (likely) desired behavior:

````
let elapsed = 0;
onLoop(({ delta }) => {
  if (!groupRef.value) return

  elapsed += delta * props.speed
  const theta = offset.value + elapsed
  groupRef.value.rotation.x = (Math.cos((theta / 4)) / 8) * props.rotationFactor
  [ ... ]
````

### Reproduction

https://stackblitz.com/edit/tresjs-basic-dpvzsu?file=src%2Fcomponents%2FTheExperience.vue

### Steps to reproduce

* Use Levioso in a TresCanvas
* Place a visible element in the Levioso
* Set speed prop
* Change speed prop

See StackBlitz.

### System Info

_No response_

### Used Package Manager

npm

### Code of Conduct

### PR DESCRIPTION
**Pull request by andretchen0**
Closes #208

## Before (`speed` scales `elapsed`):

https://github.com/Tresjs/cientos/assets/20469369/01627092-cb86-4b94-99fe-6462adf418e0

## After (`speed` scales `delta`):

https://github.com/Tresjs/cientos/assets/20469369/38646d60-1d20-4f53-b332-fe10ca391c2e

## Refactors

I took the initiative to do a few refactors, each in separate commits:

* Removing an unexposed, unchanging `ref`: `offset`
* Extracting and naming constants, and "hiding" them in a closure
* Removing some nullish checks on `props.range`

Please especially have a look at the last one:

`props.range` is assigned a default value. It seems to me that it can't be `null` or `undefined`. It's typed as `[number, number]`, so the type checker should signal an error, e.g., if `:range="[1]"`.

The other `props` fields don't have null checks.

So I removed the null checks on `props.range`, but maybe I'm missing something.

Thanks!

### DIFF
--- a/src/core/abstractions/Levioso.vue
+++ b/src/core/abstractions/Levioso.vue
@@ -27,15 +27,16 @@ defineExpose({
 const { onLoop } = useRenderLoop()
 
 const offset = ref(Math.random() * 10000)
-
-onLoop(({ elapsed }) => {
+let elapsed = 0
+onLoop(({ delta }) => {
   if (!groupRef.value) return
 
+  elapsed += delta * props.speed * 0.25
   const theta = offset.value + elapsed
-  groupRef.value.rotation.x = (Math.cos((theta / 4) * props.speed) / 8) * props.rotationFactor
-  groupRef.value.rotation.y = (Math.sin((theta / 4) * props.speed) / 8) * props.rotationFactor
-  groupRef.value.rotation.z = (Math.sin((theta / 4) * props.speed) / 20) * props.rotationFactor
-  let yPosition = Math.sin((theta / 4) * props.speed) / 10
+  groupRef.value.rotation.x = (Math.cos(theta) / 8) * props.rotationFactor
+  groupRef.value.rotation.y = (Math.sin(theta) / 8) * props.rotationFactor
+  groupRef.value.rotation.z = (Math.sin(theta) / 20) * props.rotationFactor
+  let yPosition = Math.sin(theta) / 10
   yPosition = MathUtils.mapLinear(yPosition, -0.1, 0.1, props.range?.[0] ?? -0.1, props.range?.[1] ?? 0.1)
   groupRef.value.position.y = yPosition * props.floatFactor
 })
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by andretchen0** [comment]
Hey, @alvarosabu 

Any chance this will get merged?

My concrete use case:

I used Levioso in a demo for Lensflare with Leches controls.

In order to show off the Lensflare, it has to move behind other elements, but that's not good for seeing the effects of modifying props.

So, when the Leches controls modify props, I slow the Levioso to a stop for a moment. That way the user can see the effects of the new props settings.

But changing the Levioso speed produces the jerky movement at the top of the PR.

**Comment by alvarosabu** [comment]
@andretchen0 sorry it took too long for me to take a look, I'm gonna check it

File: src/core/abstractions/Levioso.vue [hunk 1]
```diff
@@ -24,21 +24,29 @@ defineExpose({
...
+  const AMPLITUDE_ROTATION_Z = 1 / 20
+  const START_OFFSET = Math.random() * 10000
+
+  const { onLoop } = useRenderLoop()
+  let elapsed = START_OFFSET
+
+  onLoop(({ delta }) => {
+    if (!groupRef.value) return
+
+    elapsed += delta * props.speed
```
alvarosabu, line 40: We could deconstruct props with toRefs if you want.

```
const { rotationFactor, range, floatFactor } = toRefs(props)
```

alvarosabu (approved):

PR #214, opened 2023-09-15
feat(cientos): 178 v-light-helper

JaimeTorrealba:
This will not close the issue 178

### DIFF
--- a/docs/guide/directives/v-light-helper.md
+++ b/docs/guide/directives/v-light-helper.md
@@ -1,0 +1,34 @@
+# v-light-helper 🔆
+
+With the new directive v-light-helper provided by **TresJS**, you can add fast the respective helper to your lights with just one line of code 😍.
+
+```vue{3}
+<script setup lang="ts">
+import { OrbitControls, Sphere, vLightHelper } from '@tresjs/cientos'
+</script>
+<template>
+    <TresCanvas >
+      <TresPerspectiveCamera :position="[0, 2, 5]" />
+      <TresDirectionalLight
+        v-light-helper
+      />
+      <TresPointLight
+        v-light-helper
+      />
+      <TresSpotLight
+        v-light-helper
+      />
+
+      <TresHemisphereLight
+        v-light-helper
+      />
+  </TresCanvas>
+</template>
+```
+
+Note that you can pass a modifier with the name of a property, for example `v-log:material`, and will log directly the `material` property 
+
+::: warning
+This directive just work with the following lights:DirectionalLight,PointLight, SpotLight, HemisphereLight.
+By this way you can't tweaks the props for the helper if you need to do that, please use the normal helper instance
+:::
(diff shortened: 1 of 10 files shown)


src/core/directives/vLightHelper.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const { logWarning } = useLogger()
+
+export const vLightHelper = {
+  mounted: (el: any) => {
+    if (!el.isLight) {
+      logWarning(`${el.type} is not a light`)
+      return
+    }
+    const currentHelper = helpers[el.type]
```
alvarosabu, line 12: Looks good, add a unmounted to dispose the helper as well

alvarosabu (changes requested):

src/core/directives/vLightHelper.ts
```diff
@@ -12,6 +12,15 @@ export const vLightHelper = {
...
     el.parent.add(new currentHelper(el))
   },
+  unmounted: (el: any) => {
+    if (!el.isLight) {
+      logWarning(`${el.type} is not a light`)
+      return
+    }
+    const currentHelper = helpers[el.type]
+    const currentInstance = el.parent.children.find((child: any) => child instanceof currentHelper)
+    el.parent.remove(currentInstance)
```
alvarosabu, line 22: @JaimeTorrealba  The helper would not be free from memory unless you call `.dispose()` 
Screenshot 2023-09-16 at 17 04 44

https://threejs.org/docs/#manual/en/introduction/How-to-dispose-of-objects

alvarosabu (changes requested):

src/core/directives/vLightHelper.ts
JaimeTorrealba: Good eye there, thanks, I forgot to dispose it.

Fix is on his way :)

alvarosabu (approved):

docs/guide/directives/v-light-helper.md
```diff
@@ -0,0 +1,34 @@
...
+      />
+
+      <TresHemisphereLight
+        v-light-helper
+      />
+  </TresCanvas>
+</template>
+```
+
+Note that you can pass a modifier with the name of a property, for example `v-log:material`, and will log directly the `material` property
```
alvarosabu, line 29: Copy paste ;) @JaimeTorrealba

alvarosabu (changes requested):

alvarosabu:
@JaimeTorrealba dunno if it's possible but add an update call `helper.update()` everytime there is a change in the light. Otherwise the helped will remind in the same place if you move the light or change the color

JaimeTorrealba:
Interesting, I could use it with the renderLoop, or in the update directive cycle

1. With the renderloop is the same as the request animation frame, will not be performance problems right?
2. Option 2 it will only trigger when a reactive property (Vue) of the element changes (which is not always the case)

alvarosabu:
> Interesting, I could use it with the renderLoop, or in the update directive cycle
> 
> 1. With the renderloop is the same as the request animation frame, will not be performance problems right?
> 2. Option 2 it will only trigger when a reactive property (Vue) of the element changes (which is not always the case)

I will go for the second option since the update is not like the `controls.update()` which needs to be inside of the renderLoop

alvarosabu (approved):

PR #215, opened 2023-09-17
feat: add index routes pages

JaimeTorrealba:
Little PR to add the routes in the index page

(the CSS is very simple, I'm not good using UnoCSS) Any help appreciated


JaimeTorrealba:
@andretchen0 would you like to tweak a little, this one? Basically just add some styles?

andretchen0:
@JaimeTorrealba 

Sure thing. I don't know much about Uno either, but I'll see what I can do.

andretchen0:
> Basically just add some styles?

Since you mentioned Uno:

Looking at the playground directory, Uno isn't included and it's not a Vitepress site. That means we can't easily the same styles from the docs site, as far as I can tell.

* Do we want to add Uno or VitePress? 
* Or should we just add some plain CSS styles?

JaimeTorrealba:
Hi, For this issue please be creative, what I did just work, but is ... ugly hahaha

I don't understand well how this work, Uno seems not being installed but some "tailwinish" styles are available (I don't work with tailwind) dunno what configuration did Alvaro regarding the styles.

Anyway plain CSS is fine for me, this is just for improving a little the DX

andretchen0:
Looks like there was a version of Uno in there. I used Tailwind-type styles to format the landing page.

There was a conflict with the styles required by full-page THREE components. The styles in /playground/src/style.css made the landing page always show a scrollbar. So I created an override using the route name "Home" as a class on the body.

While I was at it, I added the route name to the document title (the one shown in the browser chrome).

andretchen0 (approved):

JaimeTorrealba:
I love the final result, thanks for this @andretchen0

PR #219, opened 2023-09-19
fix: 147 enhance environment and useenvironment abstractions

### LINKED ISSUE #147: Enhance Environment and useEnvironment abstractions
### Description

As a developer using TresJS, I want 
- Environment props to be reactive
- More presets other than 'sunset' (city, etc)

 so that:
- I can dynamically configure the preset or toggle the `background` property

### Suggested solution

Modify the `useEnvironment/component.ts` component so:

- Convert to vue file SFC
-  texture should be a ref
- Set the default for the props here.

Modify the `useEnvironment/index.ts` composable so:

- Reacts to prop changes (Using watchers and computed)
- Return the texture as a ref

 

### Alternative

_No response_

### Additional context

_No response_

### Validations

alvarosabu:
Closes #147


docs/.vitepress/theme/components/EnvironmentDemo.vue
```diff
@@ -0,0 +1,92 @@
...
+    step: 0.01,
+  },
+}, {
+  uuid: 'environment',
+})
+
+const environmentRef = ref(null)
+
+watchEffect(() => {
+  console.log(background.value.value)
```
JaimeTorrealba, line 36: Let's remove the console in the docs,

JaimeTorrealba (changes requested):
I like the component, just remove a bunch of unnecessary stuff, and merge with main :)

Thanks for this man

PR #220, opened 2023-09-19
feat: 178 v distance to

### LINKED ISSUE #178: V-directives in cientos proposal
### Description

After some investigation, we can add a new section to the Cientos pkg regarding the Vue custom directives:

The main goal of this would be to provide more DX for the Tresjs Ecosystem:

We can't install them globally, because that will interfere with the treeshaking, so they will be imported and used within each component

Some of the initial proposals right now are:
- [x] **v-log**:  to log the mesh, material/geometry, etc. within the component without making a reference and watching it
- [x] **v-look-at**:  as we can use the onRenderLoop inside the directive, we could reference a moving object so that it will always update the look-at property
- [ ] **v-responsive**: a proposal that will accept a breakpoint and apply a function to the current mesh
- [ ] **v-follow**: a simple way to follow another mesh, maintaining the initial distance
- [ ] **v-distance-to**: a simple way to calculate the distance 
- [x] **v-light-helper**: create helper for the provided light

### Suggested solution

EJ.
const vLog= {
  mounted: (el) => console.log(el)
}
...

    <TresMesh v-log>
      <TresTorusGeometry />
      <TresMeshBasicMaterial />
    </TresMesh>

### Alternative

Any other directive that you can think of will be appreciated

### Additional context

_No response_

### Validations

JaimeTorrealba:
closes #178


JaimeTorrealba:
@andretchen0 I know if a lot to see, but if you find time, without rush please check this too

src/core/directives/vDistanceTo.ts
```diff
@@ -0,0 +1,29 @@
...
+    const observer = extractBindingPosition(binding)
+    if (arrowHelper) {
+      arrowHelper.dispose()
+      el.parent.remove(arrowHelper) 
+    }
+    const dir = observer.clone().sub(el.position)
+    dir.normalize()
+    arrowHelper = new ArrowHelper( dir, el.position, el.position.distanceTo(observer) / 1.5, 0xffff00 )
+    el.parent.add( arrowHelper )
+    // TODO see other way to show the distance
```
andretchen0, line 15: Keeping the console as the default doesn't bug me here – I'm guessing `v-distance-to` is mainly a debugging tool, used by people who know how things are logged to the console.

For other ideas, [you might have a look here](https://vuejs.org/guide/reusability/custom-directives.html#hook-arguments) for how to pass extra args.

E.g. StatsGL uses a 'mode' field. If we want to stick with that, then maybe something like ...:

```
// Default: log to console
<Sphere v-distance-to="otherSphereRef">
```

```
// Display in a custom StatsGL window
<Sphere v-distance-to:mode.statsgl="otherSphereRef">
```

playground/src/pages/directives/vDistanceTo.vue
```diff
@@ -0,0 +1,44 @@
+<script setup lang="ts">
+import { shallowRef } from 'vue'
+import { TresCanvas, useRenderLoop } from '@tresjs/core'
```
andretchen0, line 3: `useRenderLoop` is unused, so it can be removed.

src/core/directives/vDistanceTo.ts
```diff
@@ -0,0 +1,29 @@
+import { ArrowHelper } from 'three'
+import { extractBindingPosition } from '../../utils/index'
+
+export const vDistanceTo = {
+  updated: (el: any, binding: any) => {
+    const observer = extractBindingPosition(binding)
```
andretchen0, line 6: Currently, if you pass an empty ref to `v-distance-to` it throws an error, stopping the app.

Maybe it would be useful to fail silently or `console.warn` here if the passed ref doesn't point to an object with a position.

I tried switching between two objects with a single ref. This works as expected:

```
<script setup lang="ts">
import { shallowRef } from 'vue'
import { TresCanvas, useRenderLoop } from '@tresjs/core'
import { OrbitControls, Sphere, vDistanceTo } from '@tresjs/cientos'
import { SRGBColorSpace, NoToneMapping, Vector3 } from 'three'
import { useControls, TresLeches } from '@tresjs/leches'
import '@tresjs/leches/styles'

const gl = {
  clearColor: '#333',
  outputColorSpace: SRGBColorSpace,
  toneMapping: NoToneMapping,
}

const sphere1Ref = shallowRef()
const sphere2Ref = shallowRef()
const distToRef = shallowRef({clone:() => new Vector3(0,0,0), position:new Vector3(0,0,0)})

const { value: slider } = useControls({
  slider: {
    value: -2,
    min: -4,
    max: 2,
    step: 0.1,
  },
})

useRenderLoop().onLoop(({elapsed}) => {
  if (Math.sin(elapsed) > 0) {
    distToRef.value = sphere1Ref.value
  } else {
    distToRef.value = sphere2Ref.value
  }
})
</script>

<template>
  <TresLeches class="top-0 important-left-4" />
  <TresCanvas v-bind="gl">
    <TresPerspectiveCamera :position="[0, 2, 5]" />
    <Sphere
      ref="sphere1Ref"
      :position="[-2, slider, 0]"
      :scale="0.5"
    />
    <Sphere
      ref="sphere2Ref"
      :position="[5, slider, 0]"
      :scale="0.5"
    />
    <Sphere
      v-distance-to="distToRef"
      :position="[2, 0, 0]"
      :scale="0.5"
    />
    <TresGridHelper :args="[10, 10]" />
    <OrbitControls />
  </TresCanvas>
</template>
```

So maybe it would be handy if this ...

```
const distToRef = shallowRef({clone:() => new Vector3(0,0,0), position:new Vector3(0,0,0)})
```

... could just be this momentarily, without causing an error ...

```
const distToRef = shallowRef()
```

andretchen0 (changes requested):
Works as advertised. 👍

There a couple language changes to make in the docs. Nothing major.

Otherwise, I made some suggestions:

* Based on the comments in the code, you were looking for a way to incorporate other displays. I made a suggestion for the API, but feel free to implement or not.
* Currently, passing a null ref to `v-distance-to` throws an unhandled error. Personally, I'd prefer to have a guard there to do some type checking, then `console.warn` if the ref isn't an object with a position. But it's your call.

docs/guide/directives/v-distance-to.md
```diff
@@ -0,0 +1,36 @@
+# v-distance-to
+
+¿Have you tried to calculate the distance between two object with ThreeJs?
```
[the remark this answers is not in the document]
JaimeTorrealba: This is because I'm Spanish speaker hahaha

src/core/directives/vDistanceTo.ts
JaimeTorrealba: For now, I'll just let the consoles, if the people use this directive, and it's necessary in the future we could improve this one

src/core/directives/vDistanceTo.ts
JaimeTorrealba: I handle the error with a simple validation and logging the warm, This was also the case for the vAlwaysLookAt directive. Thanks for detect this

andretchen0: Great! That keeps the app from stopping.

Can I make a further suggestion?

Can you add some info to the logged message? Maybe ...

``
logWarning(`v-distance-to: ${binding} is not a "Object3D"`)
``

## Context

Here's what `useLogger` currently outputs:

Screenshot 2023-10-06 at 16 30 18

* Unlike with a thrown `Error`, when `console.log`ging, you don't get a call stack. You just get the file/line number of the `console.log`. 
* And the file/line number reported by `useLogger` is `useLogger`'s own line number, not `v-distance-to`'s.

That means you don't get context for the problem, making it hard to find.

andretchen0 (changes requested):
Looks good!

Just one more thing! 🤪

JaimeTorrealba:
If I was better with Typescript I could definitely do something better, but for now I'll validate manually :D

I appreciate this feedbacks

JaimeTorrealba:
@andretchen0 I'll be on vacations from 14 - oct, to 04 nov

So I'll not make more PRs for now, please take care of cientos 🙏

andretchen0 (approved):
Looks good!

andretchen0:
> @andretchen0 I'll be on vacations from 14 - oct, to 04 nov

Sounds good. See you in november!

alvarosabu:
@andretchen0 Hey there! Thanks for taking care of reviewing, what do you think, ready to merge?

alvarosabu (approved):
Looks pretty handy @JaimeTorrealba

andretchen0:
> @andretchen0 Hey there! Thanks for taking care of reviewing, what do you think, ready to merge?

It's ready to go, I believe.

Issue #222, opened 2023-09-20
catmullromline 


JaimeTorrealba:
### Description

As a developer using Tresjs I would like to add catmullromline components, in shapes

<CatmullRomLine
  points={[[0, 0, 0], ...]}      
  closed={false}                 
  curveType="centripetal"         
  tension={0.5}                  
  color="black"                   
  lineWidth={1}                   
  dashed={false}                  
  vertexColors={[[0, 0, 0], ...]} 
  {...lineProps}                 
  {...materialProps}              
/>

### Suggested solution

Similar to: [catmullromline](https://github.com/pmndrs/drei#catmullromline)

### Alternative

Any other alternative is considered

### Validations

- [X] I agree to follow this project's [Code of Conduct](https://github.com/Tresjs/cientos/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/Tresjs/cientos/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://cientos.tresjs.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/tresjs/cientos/issues) that reports the same bug to avoid creating a duplicate.

JaimeTorrealba:
@andretchen0 we could close this issue right?

andretchen0:
> @andretchen0 we could close this issue right?

Yes

For what it's worth, I can close issues, merge, etc., but I prefer to leave those decisions to you guys.

JaimeTorrealba:
They close automatically if we add close "#NUMBER-ISSUE" in the PR :)

andretchen0:
> They close automatically if we add close "#NUMBER-ISSUE" in the PR :)

😀

PR #239, opened 2023-09-30
feat(Lines): add CatmullRomCurve3, Line2

andretchen0:
#222

This PR adds two new components:

* CatmullRomCurve3
* Line2

As a side note, I simply adopted the Three.js names. "Line2" isn't a 2-D line; I think it's just an alternative version of Three.js's first "Line". 

Previews testing reactivity are available after `pnpm run playground` under "Shapes".

Docs are available in the deploy preview below, also under "Shapes".

## Reactivity/Performance/Flexibility

I chose to make everything reactive. And I chose to make it as flexible as possible, following Drei's example. That comes at a performance cost though.

### Vertex colors

The prop `colors` optionally takes vertex colors in the form of `TresColor[]`. Using `TresColor` is flexible, but it means that the colors have to be normalized before they can be used. And if the user updates `colors` every frame, the entries have to be normalized every frame.

Following the Drei implementation, `colors` are interpolated. That means that users don't have to supply a color for each vertex, the supplied colors will instead be lerped as needed. But that's an extra operation that wouldn't normally need to be done.

To keep code simple, the interpolation is done every time the colors and/or points are updated – even if no points were added/removed.

### Points

The Line2 `points` prop has the following type:

`Array<Vector3 | Vector2 | [number, number, number] | [number, number] | number>`

allowing different data types in the same array. That also has to be normalized whenever points or colors are updated.

## Background

The Drei version of CamullRomCurve divides the path into n segments. Then *those* points are forwarded to a Drei `<Line>`. 

I took the same approach here, but afaik, there's no `Line` built in to Tres or Cientos that does what Drei's `<Line />` does. So I added one.

Three's "Line2" allows the user to specify line thickness. "Line" has the option, but the implementation relies on local WebGL, it appears – it doesn't work on most platforms.


playground/src/pages/shapes/CatmullRomCurve3Demo.vue
```diff
@@ -0,0 +1,86 @@
...
+  [0.8, 0.0, 0.2],
+  [0.9, 0.0, 0.1],
+])
+
+const lineWidth = ref(5)
+const dashed = ref(true)
+const dashSize = ref(1)
+const dashScaleRef = ref(50)
+const dashOffset = ref(0)
+const gapSize = ref(1)
```
JaimeTorrealba, line 49: You're now using it :)

playground/src/pages/shapes/Line2Demo.vue
```diff
@@ -0,0 +1,79 @@
+<script setup lang="ts">
+import { TresCanvas, useRenderLoop } from '@tresjs/core'
+import { OrbitControls, Line2 } from '@tresjs/cientos'
+
+const gl = {
```
JaimeTorrealba, line 5: Same here, you're not using this gl

JaimeTorrealba (commented):
I seem everything fine with this 👍 

Next time, please create a new issue for Line2 and treat like different PR (is better for the changelog and the final releases notes. Now is fine, let's finish it and merge them

Something call my attention but is not in your PR, the OrbitControls is having problems, I don't want to merge it before knowing what is, but when is ready I'll approve it

andretchen0:
Good catches. Thanks.

> Next time, please create a new issue for Line2 and treat like different PR (is better for the changelog and the final releases notes. Now is fine, let's finish it and merge them

It's no problem to split it up. If you'd like me to do that, just let me know.

> Something call my attention but is not in your PR, the OrbitControls is having problems, I don't want to merge it before knowing what is, but when is ready I'll approve it

Is there something I can help with?

JaimeTorrealba:
No problem, let's merge them like so, no need to split.

The second problem maybe is me, you don't have that issue?

andretchen0:
> The second problem maybe is me, you don't have that issue?

I'm not sure exactly what you're referring to, but I don't notice a problem with OrbitCintrols.

JaimeTorrealba (approved):

PR #241, opened 2023-10-04
feat: 209 global audio

### LINKED ISSUE #209: Global audio component
### Description

As a developer using Tresjs I would like to have a simple way to add a global sound.

### Suggested solution

An abstraction of: https://threejs.org/docs/index.html?q=audio#api/en/audio/Audio

You could consider something like:
<Audio src="MYAUDIO_PATH.mp3" :volume="0.3" /> 

As example

### Alternative

Feel free to suggest other alternatives

### Additional context

_No response_

### Validations

JaimeTorrealba:
closes #209


docs/guide/abstractions/global-audio.md
```diff
@@ -0,0 +1,39 @@
...
+  </TresCanvas>
+</template>
+```
+
+## Props
+
+| Prop           | Description                                        | Default               |
+| :------------- | :------------------------------------------------- | --------------------- |
+| `src`          | Path to your audio file                            |                       |
+| `playElement`  | Id of the DOM element that trigger the play state  | `renderer.domElement` |
```
andretchen0, line 28: > DOM element that trigger

triggers

On the following lines as well.

docs/guide/abstractions/global-audio.md
```diff
@@ -0,0 +1,39 @@
...
+| `pauseElement` | Id of the DOM element that trigger the pause state |                       |
+| `stopElement`  | Id of the DOM element that trigger the stop state  |                       |
+| `loop`         | If the audio must be replayed when ends            | `false`               |
+| `volume`       | Volume of the audio.                               | `0.5`                 |
```
andretchen0, line 32: > Volume of the audio.

No period to match other descriptions.

playground/src/pages/abstractions/GlobalAudioDemo.vue
```diff
@@ -0,0 +1,66 @@
...
+  <TresCanvas v-bind="gl">
+    <TresPerspectiveCamera
+      :position="[0, 0, 7.5]"
+      :fov="75"
+      :near="0.1"
+      :far="1000"
+    />
+    <GlobalAudio
+      :src="exampleAudio"
+      :volume="options.volume.value.volume"
```
andretchen0, line 42: I tried to run the playground demo, and the following error was thrown:

```
Uncaught (in promise) TypeError: $setup.options.volume is undefined
    default GlobalAudioDemo.vue:41
    renderFnWithContext runtime-core.esm-bundler.js:766
    setup tres.js:558
    renderComponentRoot runtime-core.esm-bundler.js:816
    componentUpdateFn runtime-core.esm-bundler.js:5701
    run reactivity.esm-bundler.js:178
    update runtime-core.esm-bundler.js:5814
    setupRenderEffect runtime-core.esm-bundler.js:5822
    mountComponent runtime-core.esm-bundler.js:5612
    processComponent runtime-core.esm-bundler.js:5565
```

andretchen0 (changes requested):
Hey @JaimeTorrealba , the demo didn't work for me, so I couldn't try it out. I marked the line that the error sent me to and made some other remarks.

Once that's fixed, rerequest a review and I'll have another look. 🙂

src/core/abstractions/GlobalAudio.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    expose({ sound })
+
+    onUnmounted(() => {
+      if (sound) {
+        sound.disconnect()
+      }
+    })
+
+    watch(() => [props.playbackRate], () => sound.setPlaybackRate( props.playbackRate || 1 ), { immediate: true })
```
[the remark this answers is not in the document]
JaimeTorrealba: Good eye, thanks in this case is better to use it

playground/src/pages/abstractions/GlobalAudioDemo.vue
JaimeTorrealba: Could be the version of Leches?? I don't have that problem, I'll hardcode the volume param, please let me know if after this you can use the demo

src/core/abstractions/GlobalAudio.ts
```diff
@@ -0,0 +1,128 @@
...
+    
+    const selector = document.getElementById(props.playElement ?? '')
+    const btnPlay = selector ? selector : renderer.value.domElement
+    useEventListener(btnPlay, 'click', () => {
+      sound.play()
+      emit('isPlaying', sound.isPlaying)
+    })
+
+    const btnPause = document.getElementById(props.pauseElement ?? '')
+    if (btnPause) {
```
andretchen0, line 112: I don't know if it can be worked in, but I was a little confused at the beginning when interacting with the play/pause functionality of the demo.

I was expecting:

* hit pause - audio pauses
* hit pause - audio plays
* hit pause - audio pauses

Instead, this happens:

* hit pause - audio pauses
* hit pause - no change
* hit pause - no change

So, the would work sort of like the [standard HTML audio widget](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio).

I don't know if that functionality can be worked in or if it's even desirable, but I had a moment of confusion there, so I thought that might be useful to pass on.

andretchen0 (approved):

src/core/abstractions/GlobalAudio.ts
JaimeTorrealba: But what you say make sense, and we can remove the "pauseElement" prop.

Yeah, I like the idea, but I need to do some changes. I will work on these thanks for this great feedback

PR #245, opened 2023-10-05
docs: import demos

### LINKED ISSUE #242: Docs: use imported code snippets where possible
### Description

Code snippets in docs sometimes [contain typos](https://github.com/Tresjs/tres/pull/411) or other errors.

### Suggested solution

Where possible, use [Vitepress' markdown plugin](https://vitepress.dev/guide/markdown#import-code-snippets) to import demo code snippets directly, rather than copy/pasting them:

```
<<< @/filepath{highlightLines}
```

### Alternative

_No response_

### Additional context

@JaimeTorrealba suggested standardizing this approach after I [submitted this PR](https://github.com/Tresjs/cientos/pull/240) using the plugin as outlined.

### Validations

andretchen0:
This closes #242 .

* Since the demo source wasn't written to be viewed directly, there was often more code in them then absolutely necessary. I tried to take as much out as I could while maintaining the look of the demo – e.g., removing `<TresCanvas v-bind="gl" />` and `gl` object and replacing it with `<TresCanvas clear-color="#333">`, or removing refs. I don't think that's noticeable for the most part, but let me know if I went too far.
* In a few places, I couldn't shorten the demos and maintain the visual effects or functionality, so I left the copy/pasted source code snippets. `BackdropDemo.vue` is one instance. (Note that it *is* possible to include a [VSCode-style region](https://vitepress.dev/guide/markdown#import-code-snippets), but I didn't want to complicate things. Only whole files are imported at the moment.)
* Ideally, code highlighting would be specified in the`.vue` demo files, not in the `.md` files. That's possible with the current system. But when the `.vue` files are parsed for rendering, they throw an error if used *within* a Vue component, e.g.:

```
<Box
   color="orange" // [!code hl]
   :size="1"
/>
```

Highlighting a component over multiple lines is a really common case for us. Since that breaks the demos, to keep things consistent, I left all highlighting in the `.md` files. That looks like this:

```
<<< @/.vitepress/theme/components/HtmlOccludeDemo.vue{2,6,19,28}
```

Above the lines 2, 6, 19, and 28 are highlighted.

### DIFF
--- a/docs/components.d.ts
+++ b/docs/components.d.ts
@@ -8,11 +8,9 @@ export {}
 declare module 'vue' {
   export interface GlobalComponents {
     BackdropDemo: typeof import('./.vitepress/theme/components/BackdropDemo.vue')['default']
-    BackdropSimpleDemo: typeof import('./.vitepress/theme/components/BackdropSimpleDemo.vue')['default']
     CameraControlsDemo: typeof import('./.vitepress/theme/components/CameraControlsDemo.vue')['default']
     CatmullRomCurve3Demo: typeof import('./.vitepress/theme/components/CatmullRomCurve3Demo.vue')['default']
     ContactShadowDemo: typeof import('./.vitepress/theme/components/ContactShadowDemo.vue')['default']
-    copy: typeof import('./.vitepress/theme/components/FBXModelDemo copy.vue')['default']
     DocsDemo: typeof import('./.vitepress/theme/components/DocsDemo.vue')['default']
     FBXModelDemo: typeof import('./.vitepress/theme/components/FBXModelDemo.vue')['default']
     Feather: typeof import('./.vitepress/theme/components/Feather.vue')['default']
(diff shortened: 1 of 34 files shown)


docs/components.d.ts
```diff
@@ -8,9 +8,11 @@ export {}
 declare module 'vue' {
   export interface GlobalComponents {
     BackdropDemo: typeof import('./.vitepress/theme/components/BackdropDemo.vue')['default']
+    BackdropSimpleDemo: typeof import('./.vitepress/theme/components/BackdropSimpleDemo.vue')['default']
```
JaimeTorrealba, line 11: Let's try to never update this files, sometimes gives us unnecessary conflicts

JaimeTorrealba (approved):
Thank you so much for this PR, solutions like this help us to scale more

I know is not specially fun but avoid a lot of typos problems

🙏🏼

JaimeTorrealba:
And don't worry about the moment, we can use the  '<<< @/.vitepress' Format.

This is already a great DX change for us. We could always go back to this later

JaimeTorrealba:
I approved already, but before merge, let me know when you delete the .d.ts file please

docs/components.d.ts
andretchen0: I've reset it to the current main branch's version and pushed.

I usually remember to reset it by hand, but forgot this time. It's automatically generated, isn't it? Do you know if we can add it to .gitignore?

andretchen0:
> Thank you so much for this PR, solutions like this help us to scale more
> 
> I know is not specially fun but avoid a lot of typos problems

I don't mind. It's better than trying to keep a copy of the demos in sync.

The next big missing piece for me is the props section of the docs. It'd really be great if those tables could be pulled from the components. I think most of the component have JSDocs for the props.

Ideas are welcome! 😀

JaimeTorrealba:
Yes, as you can see we're a small numbers of maintainers here, a cientos was basically maintained just by me (Alvaro helps a lot, but he has other pkgs too, and Tino is full focus on the post-processing  pkg)

So your addition is really appreciated, we could talk about how improve the pkg and the DX (not only do components)

Also, about another pkg for the ecosystem, I send to you a discussion in Discord about other future pkg 

If you want, I'm available for discussing more on discord :)

PR #274, opened 2023-10-30
feat(Superformula): add superformula

### LINKED ISSUE #270: New shape: Superformula
### Description

As a developer using Tres/Cientos, it'd be nice to have a 3D [Superformula](https://en.wikipedia.org/wiki/Superformula) shape available in Cientos.

### Suggested solution

Something like this:

https://github.com/Tresjs/cientos/assets/20469369/46201394-6fed-42d9-ac25-83cce9903c37

### Alternative

_No response_

### Additional context

_No response_

### Validations

andretchen0:
Closes #270

* Run `pnpm run docs:dev` and see `/guide/shapes/superformula.html`.
* Run `pnpm run playground` and see `/shapes/superformula`.

NOTE: Uses the unmerged `.gitignore` from [this PR.](https://github.com/Tresjs/cientos/pull/272)

### DIFF
--- a/src/core/shapes/Superformula.vue
+++ b/src/core/shapes/Superformula.vue
@@ -1,0 +1,185 @@
+<script setup lang="ts">
+import { BufferAttribute, BufferGeometry } from 'three'
+import { shallowRef, watch } from 'vue'
+import type { TresColor } from '@tresjs/core'
+import { onUnmounted } from 'vue'
+
+export interface SuperFormulaProps {
+  /**
+     * Number of horizontal mesh segments
+     */
+  widthSegments?: number
+  /**
+     * Number of vertical mesh segments
+     */
+  heightSegments?: number
+  /**
+     * The 3D Superformula is the spherical product of 2 2D superformula curves: here called curves "A" and "B".
+     * Number of radial arms/ripples of A, corresponding to "m" [in this article.](https://en.wikipedia.org/wiki/Superformula)
+     */
+  numArmsA?: number
+  /**
+     * A's first exponent
+     */
+  expA1?: number
+  /**
+     * A's second exponent
+     */
+  expA2?: number
+  /**
+     * A's third exponent
+     */
+  expA3?: number
+  /**
+     * For B, number of radial arms/ripples
+     */
+  numArmsB?: number
+  /**
+     * B's first exponent
+     */
+  expB1?: number
+  /**
+     * B's second exponent
+     */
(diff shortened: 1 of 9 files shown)


src/core/shapes/Superformula.vue
```diff
@@ -0,0 +1,184 @@
...
+     */
+  expA3?: number
+  /**
+     * For B, number of radial arms/ripples
+     */
+  numArmsB?: number
+  /**
+     * B's first exponent
+     */
+  expB1?: number
```
alvarosabu, line 39: @andretchen0 looks amazing, I think I haven't seen this since university haha. When animated looks mesmerizing.

Wonder if this is the best way we can over the props: 

The exponents are limited to 3 or to "N"? Could the user pass them as arrays?

andretchen0: > @andretchen0 looks amazing, I think I haven't seen this since university haha. When animated looks mesmerizing.
> 
> Wonder if this is the best way we can over the props: 
> 
> The exponents are limited to 3 or to "N"? 

They are limited to 3. (The 3d plot is the product of 2 2d plots. Each 2d plot has 3 exponents.)

> Could the user pass them as arrays?

Do you want to offer both `:args` *and* `:exp-a1`, etc?

Fwiw, I started out using positional arguments, but switched to named arguments. All the args are optional and there are currently 8 of them. In the case of positional args, that leads to calls like:

```
<Superformula :args="[4, undefined, undefined, undefined, 3, undefined, undefined, 1]" />
```

Intellisense helps out, but it still seems kind of hairy.

Three's SphereGeometry also offers

phiStart : Float, phiLength : Float, 			thetaStart : Float, thetaLength : Float) 	

... Which could be handy. With those you can draw, say, a quarter of the shape. We could use that to make a Cientos heart from 2 <Superformula>

But it means even more args ...

What do you think?

alvarosabu: Hey there @andretchen0 , normally `:args` is reserved for the class constructor params. I was thinking more of:

```
<Superformula :exp-b="[b1, b2, b3]" />
```

andretchen0:
```
<Superformula :exp-b="[b1, b2, b3]" />
```

Ah, I see. Sure, we can do that.

alvarosabu (approved):
Amazing job @andretchen0 Thanks for grouping the props, lets merge this

PR #288, opened 2023-11-21
fix: initialize progress value with init scroll

alvarosabu:
This PR fixes the issue with the progress value being always 0 even if the page is refreshed after scrolling to a certain position. (Astro)

### DIFF
--- a/src/core/controls/ScrollControls.vue
+++ b/src/core/controls/ScrollControls.vue
@@ -82,7 +82,6 @@ const { height, width } = useWindowSize()
 let initCameraPos = 0
 const initialized = ref(false)
 
-const progress = ref(0)
 const progressScroll = ref(0)
 const scrollNodeY = ref(0)
 const direction = props.horizontal ? 'x' : 'y'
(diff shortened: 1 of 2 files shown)


src/core/controls/ScrollControls.vue
```diff
@@ -112,21 +111,21 @@ watch(
   },
 )
 
-watch(windowY, (value) => {
-  if (!isScrolling.value && !props.htmlScroll) return
+function updateScroll(value: number) {
   progressScroll.value = (value / height.value / (scrollNodeY.value / height.value - 1))
```
JaimeTorrealba, line 115: How is this working for horizontal axis??

JaimeTorrealba (changes requested):

src/core/controls/ScrollControls.vue
alvarosabu: True

andretchen0 (approved):
It works! This is a welcome improvement.

I'm approving since the fix works, but I noticed this related bug.

## Bug: linking to named anchors

### Reproduction

#### Example markup

```
<a href="#two">Jump to section 2</a>
<section><h1>Section 1</h1></section>
<section><h2 id="#two">Section 2</h2></section>
```

1. In the text editor, insert the example markup into the playground/src/pages/controls/ScrollControlsDemo.vue and save.
2. On the command line, start the playground with `pnpm run playground` if not already running.
3. In the browser, open a new, blank tab.
4. In the browser, paste `http://localhost:5173/controls/scroll-controls#two` in the address bar and hit `Enter`.

### Expected behavior

The page should scroll to the second header.

### Behavior

The page loads and remains at y-offset 0.

Once the page is loaded, clicking on "Jump to section 2" works as expected. 

Tested in Chrome, FF, Safari on Mac.

----

**Would you like this to be opened as a separate issue or worked on here?**

alvarosabu:
Hi @andretchen0 thanks for the help, yes let's open it on a separate issue to focus on anchors

PR #306, opened 2023-12-10
feat: Stats-gl v2

RenaudRohlinger:
Stats-gl now supports postprocessing and multiple render per loop. The API changed a bit so I reflected the changes here.


alvarosabu (approved):
Thanks a lot @RenaudRohlinger for helping us update this, amazing work in the v2 👏🏻

PR #324, opened 2023-12-25
feat: add BakeShadows component

### LINKED ISSUE #321: BakeShadows
### Description

I would like to have a simple way to baked my shadows using threeJs and the shadow.autoUpdate function on the render as false

### Suggested solution

To create a simple renderless doing this

### Alternative

_No response_

### Additional context

_No response_

### Validations

JaimeTorrealba:
closes #321  
How can I remove these warnings

image

### DIFF
--- a/playground/src/pages/misc/BakeShadows.vue
+++ b/playground/src/pages/misc/BakeShadows.vue
@@ -1,0 +1,53 @@
+<script setup lang="ts">
+import { CameraControls, vLightHelper, BakeShadows } from '@tresjs/cientos'
+import { TresCanvas, useRenderLoop } from '@tresjs/core'
+import { NoToneMapping, SRGBColorSpace } from 'three'
+
+const gl = {
+  clearColor: '#82DBC5',
+  outputColorSpace: SRGBColorSpace,
+  toneMapping: NoToneMapping,
+  shadows: true,
+}
+
+const cubeRef = shallowRef()
+
+const { onLoop } = useRenderLoop()
+
+onLoop(({ elapsed }) => {
+  if (cubeRef.value) {
+    cubeRef.value.rotation.y = elapsed * 0.5
+    cubeRef.value.rotation.x = elapsed * 0.5
+  }
+})
+</script>
+
+<template>
+  <TresCanvas
+    v-bind="gl"
+  >
+    <TresPerspectiveCamera :position="[0, 2, 5]" />
+    <CameraControls />
+    <BakeShadows />
+    <TresMesh
+      ref="cubeRef"
+      cast-shadow
+    >
+      <TresBoxGeometry />
+      <TresMeshStandardMaterial :color="0x00ff00" />
+    </TresMesh>
+    <TresMesh
+      receive-shadow
+      :position="[0, -2, 0]"
+      :rotation-x="-Math.PI / 2"
+    >
+      <TresPlaneGeometry :args="[5, 5]" />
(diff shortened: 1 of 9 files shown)


src/core/misc/BakeShadows.ts
alvarosabu, line 1: @JaimeTorrealba probably because you are defining any props

JaimeTorrealba: But <Stats /> has props, and the messages still appears

andretchen0:
About the color error, the `color` prop on `TresMeshStandardMaterial` expects a `THREE.Color`, so any of [the instantiation examples here](https://threejs.org/docs/#api/en/math/Color) will work, like this:

```
<TresMeshStandardMaterial :color="new Color(0x00ff00)" />
```

src/core/misc/BakeShadows.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  setup() {
+    const { renderer } = useTresContext()
+
+    watchEffect(() => {
+      renderer.value.shadowMap.autoUpdate = false
+      renderer.value.shadowMap.needsUpdate = true
+    })
+
+    return null
```
andretchen0, line 15: Remove `return null`. I believe this fixes the typing issue. Let me know if that doesn't fix it for you.

[Fwiw, here's an example with no returned value in the Vue docs.](https://vuejs.org/guide/typescript/composition-api.html#without-script-setup)

playground/src/pages/misc/BakeShadows.vue
```diff
@@ -0,0 +1,53 @@
...
+  >
+    <TresPerspectiveCamera :position="[0, 2, 5]" />
+    <CameraControls />
+    <BakeShadows />
+    <TresMesh
+      ref="cubeRef"
+      cast-shadow
+    >
+      <TresBoxGeometry />
+      <TresMeshStandardMaterial :color="0x00ff00" />
```
andretchen0, line 37: `TresMeshStandardMaterial`'s `color` prop expects a `THREE.Color`, so you'll have to instantiate one with `new Color(...)`.

playground/src/pages/misc/BakeShadows.vue
```diff
@@ -0,0 +1,53 @@
...
+    </TresMesh>
+    <TresMesh
+      receive-shadow
+      :position="[0, -2, 0]"
+      :rotation-x="-Math.PI / 2"
+    >
+      <TresPlaneGeometry :args="[5, 5]" />
+      <TresMeshStandardMaterial :color="0xf7f7f7" />
```
andretchen0, line 45: Same as above. This should be a `THREE.Color`, not a number.

andretchen0 (changes requested):

src/core/misc/BakeShadows.ts
JaimeTorrealba: Thank you so much, it was that 🙏

playground/src/pages/misc/BakeShadows.vue
JaimeTorrealba: Same here

playground/src/pages/misc/BakeShadows.vue
JaimeTorrealba: Done :)

andretchen0 (approved):
Looks good to me!

PR #327, opened 2024-01-09
fix: Correct return type for useGLTF composable

Dotmot:
Added conditional return type to infer the actual return type when passing strings or arrays as currently it always return GTLFResult and that is not always correct.
image
With proposed changes
image
image

The type is correctly set and allows me to do this
image


JaimeTorrealba (approved):
I like it,
And thanks for fix this

PR #331, opened 2024-01-11
feat(AnimatedSprite): add AnimatedSprite

andretchen0:
@JaimeTorrealba , @alvarosabu 

I wonder if you guys would take a look at `<AnimatedSprite />` and give me some feedback. 

Docs: `pnpm run docs:dev` and then => http://localhost:5173/guide/abstractions/animated-sprite.html
Playground: `pnpm run playground` and then => http://localhost:5173/abstractions/animated-sprite

## What do you think of the API?

I'd like to know if you see any improvements to make to the API. I followed most of what Drei did, but included a few extra things that I picked up from using [Deepnight's gameBase](https://github.com/deepnight/gameBase). Maybe those changes aren't necessary here – What do you think?

In particular, these props:

* anchor – this allows differently sized images to be used in an atlas
* definitions – this allows users to specify how each animation plays: frame order and delays. I think it's really handy for tweaking animations on the fly, but it does add about 200 lines of code, mostly for the definition parser.
* animation – named animations are pulled from source image names. See guide/abstractions/animated-sprite.html#animation . I found this really handy when working with the gamebase, but maybe our users won't see the point.

## Do we want to include a spritesheet compiler?

A while ago, I wrote a very dumb node.js command line app for compiling spritesheets/atlas JSON from source images. I could rewrite it for the browser and include it in the docs. That would let users compile their own spritesheets/atlas JSON online with a drag and drop interface. But do we want to support that going forward?

TexturePacker is paid software. For context, Drei and our `<AnimatedSprite />` read simple TexturePacker atlas JSON. 
But Drei and our `<AnimatedSprite />` don't/won't support most of the advanced features of TexturePacker.

### DIFF
--- a/docs/guide/abstractions/animated-sprite.md
+++ b/docs/guide/abstractions/animated-sprite.md
@@ -1,0 +1,131 @@
+# AnimatedSprite
+
+<DocsDemo>
+  <AnimatedSpriteDemo />
+</DocsDemo>
+
+`<AnimatedSprite />` allows you to use 2D animations defined in a [texture atlas](https://en.wikipedia.org/wiki/Texture_atlas). A typical `<AnimatedSprite />` will use:
+
+* an image containing multiple sprites
+* a JSON atlas containing the coordinates of the image
+
+## Usage
+
+<<< @/.vitepress/theme/components/AnimatedSpriteDemo.vue{3,11-16}
+
+::: warning Suspense
+`<AnimatedSprite />` loads resources asynchronously, so it must be wrapped in a `<Suspense />`.
+:::
+
+## Props
+
+<CientosPropsTable 
+component-path="src/core/abstractions/AnimatedSprite/component.vue" 
+:fields="['name', 'description', 'default', 'required']"
+:on-format-value="({valueFormatted, propName, fieldName, getFieldFormatted})=> {
+  if (fieldName === 'description') {
+    let type = getFieldFormatted('type')
+    if (type.indexOf('TSFunctionType') !== -1 && propName.startsWith('on')) {
+      type = '<code>(frameName:string) => void</code>'
+    }
+    return type + ' – ' + valueFormatted
+  }
+}"
+ />
+
+## `animation`
+
+The `animation` prop holds either the name of the currently playing animation or a range of frames to play, or a frame number to display.
+
+### Named animations
+
+Frames are automatically grouped into named animations, if you use either of the following naming conventions for your source images:
+
+* `[key][frame number].[file_extension]`
+* `[key]_[frame number].[file_extension]`
+
+The `<AnimatedSprite />` will automatically make all [key] available for playback as named animations.
+
+#### Example
+
+Here are some example source image names, to be compiled into an image texture and atlas.
+
+* heroIdle00.png
+* heroIdle01.png
+* heroIdle02.png
+* heroRun00.png
+* heroRun01.png
+* heroRun02.png
+* heroRun03.png
+* heroHeal00.png
+* heroHeal01.png
+
+When the resulting image texture and atlas are provided to `<AnimatedSprite />`, "heroIdle", "heroRun", and "heroHeal" will be available as named animations. Animation names can be used as follows:
+
+```vue{3}
+<AnimatedSprite 
+atlas="..." image="..." 
+animation="heroRun" 
+/>
+```
+
+### Ranges
+
+A `[number, number]` range can be supplied as the `animation` prop. The numbers correspond to the position of the frame in the `atlas` `frames` array, starting with `0`. The first `number` in the range represents the start frame of the animation. The last `number` represents the end frame.
+
+### Single frame
+
+To display a single animation frame, a `number` can be supplied as the `animation` prop. The `number` corresponds to the position of the frame in the `atlas` `frames` array, starting with `0`. 
+
+## `anchor`
+
+The `anchor` allow you to control how differently sized source images "grow" and "shrink". Namely, they "grow out from" and "shrink towards" the anchor. `[0, 0]` places the anchor at the top left corner of the `<AnimatedSprite />`. `[1,1]` places the anchor at the bottom right corner. By default, the anchor is placed at `[0.5, 0.5]` i.e., the center.
+
+Below is a simple animation containing differently sized source images. The anchor is visible at world position `0, 0, 0`.
+<DocsDemo>
+  <AnimatedSpriteAnchorDemo />
+</DocsDemo>
+
+::: warning
+Changing the anchor from the default can have unpredictable results if `asSprite` is `true`.
+:::
+
+## `definitions`
+
+For each [named animation](#named-animations), you can supply a "definition" that specifies frame order and repeated frames (delays). For the [named animation example above](#named-animations), the `definitions` prop might look like this:
+
+```vue
+<AnimatedSprite 
+atlas="..." image="..."
+:definitions="{
+  'heroIdle':'0(5),1-2', // repeat frame 0 five times, then play frames 1-2
+  'heroRun':'0-3,2-1', // play frames 0-3, then 2-1
+  // no "heroHeal" definition; the default frame order will be used: 0-1
+}"
+/>
+```
+
+## Compiling an atlas
+
+In typical usage, `<AnimatedSprite/>` requires both the URL to a texture of compiled sprite images and a JSON atlas containing information about the sprites in the texture.
+
+* [example compiled texture](https://raw.githubusercontent.com/Tresjs/assets/6c0b087768a0a2b76148c99fc87d7e6ddc3c6d66/textures/animated-sprite/namedAnimationsTexture.png)
+* [example JSON atlas](https://raw.githubusercontent.com/Tresjs/assets/6c0b087768a0a2b76148c99fc87d7e6ddc3c6d66/textures/animated-sprite/namedAnimationsAtlas.json)
+
+Compiling source images into a texture atlas is usually handled by third-party software. You may find [TexturePacker](https://www.codeandweb.com/texturepacker) useful.
+
+## Without an atlas
+
+There may be cases where you don't want to supply a generated JSON atlas as an `atlas` prop. This is possible if you compile your source images in a single row of equally sized columns *and* set the `atlas` prop to the number of columns.
+
+### Example
+
+This image is comprised of 16 source images, compiled into a single image, in a single row:
+
+<img src="https://raw.githubusercontent.com/Tresjs/assets/6c0b087768a0a2b76148c99fc87d7e6ddc3c6d66/textures/animated-sprite/textureWithoutAtlas.png" />
+
+<DocsDemo>
+  <AnimatedSpriteNoAtlasDemo />
+</DocsDemo>
+
+<<< @/.vitepress/theme/components/AnimatedSpriteNoAtlasDemo.vue{12,13}
--- a/playground/src/pages/abstractions/AnimatedSprite.vue
+++ b/playground/src/pages/abstractions/AnimatedSprite.vue
@@ -1,0 +1,154 @@
+<script setup lang="ts">
+import { TresCanvas } from '@tresjs/core'
+import { OrbitControls, AnimatedSprite } from '@tresjs/cientos'
+import { BasicShadowMap, SRGBColorSpace, NoToneMapping } from 'three'
+import { shallowReactive } from 'vue'
+import { TresLeches, useControls } from '@tresjs/leches'
+import '@tresjs/leches/styles'
+import { TexturePackerFrameDataArray } from '../../../../src/core/abstractions/AnimatedSprite/Atlas'
+
+const gl = {
+  clearColor: '#82DBC5',
+  shadows: true,
+  alpha: false,
+  shadowMapType: BasicShadowMap,
+  outputColorSpace: SRGBColorSpace,
+  toneMapping: NoToneMapping,
+}
+
+const animationState = shallowReactive({
+  fps: 10,
+  animation: "yes",
+  flipX: false,
+  asSprite: false,
+  loop: true,
+  reversed: false,
+  resetOnEnd: false,
+  anchorX: 0.5,
+  anchorY: 0.5
+})
+
+const { fps, animation, flipX, asSprite, loop, reversed, resetOnEnd, anchorX, anchorY } = useControls({
+  fps: { value: animationState.fps, min: 0, max: 120, step: 1 },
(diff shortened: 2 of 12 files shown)


JaimeTorrealba:
Hi @andretchen0 this is a nice work, thank you so much. To be honest regarding code, almost nothing. I would like to know something.

1. If I have a spritesheet, and I would like to use only the first row, can I do that here? Without all the complicated process of the atlas, could we add that, like to be able to cut the sprite? 
I find myself using it https://lab.jaimetorrealba.com/sprites_controls_demos 

Evil Druid

Cientos Playground - AnimatedSprite

2. Should the `asSprite`  option as true by default? (what do you think?)

3. I'll put the "Compiling an atlas" section almost at the top, I believe there are people who are not familiar with it (like me) and as is "almost" required, should be good to adding right after the "usage" 

4. Non related, but did you remember how we removed the "multiple instances of ThreeJs" that blocks OrbitControls to move freely?

image

docs/guide/abstractions/animated-sprite.md
```diff
@@ -0,0 +1,131 @@
...
+When the resulting image texture and atlas are provided to `<AnimatedSprite />`, "heroIdle", "heroRun", and "heroHeal" will be available as named animations. Animation names can be used as follows:
+
+```vue{3}
+<AnimatedSprite 
+atlas="..." image="..." 
+animation="heroRun" 
+/>
+```
+
+### Ranges
```
JaimeTorrealba, line 72: Here I get confused, I tough `ranges` was a prop maybe to put, animations as range

docs/guide/abstractions/animated-sprite.md
```diff
@@ -0,0 +1,131 @@
...
+
+A `[number, number]` range can be supplied as the `animation` prop. The numbers correspond to the position of the frame in the `atlas` `frames` array, starting with `0`. The first `number` in the range represents the start frame of the animation. The last `number` represents the end frame.
+
+### Single frame
```
JaimeTorrealba, line 76: Same here :)

JaimeTorrealba (changes requested):
I know is a draft, a look really nice, I want to use this one!

andretchen0:
@JaimeTorrealba Thanks for the feedback!

andretchen0:
> If I have a spritesheet, and I would like to use only the first row, can I do that here? Without all the complicated process of the atlas, could we add that, like to be able to cut the sprite?
I find myself using it https://lab.jaimetorrealba.com/sprites_controls_demos

We can do something different, but I copied the Drei setup here. Afaik, without an atlas, the Drei version only allows for a single row of images — with all the sprites being the same size, evenly spaced, and e.g., no padding on the spritesheet itself.

We could take `[number, number]` on the atlas prop to mean number of columns and number of rows. Again with all sprites being the same size, evenly spaced, and no padding on the spritesheet itself.

That would work for the first row of your example spritesheet, I think. And that's an ok change in my book as far as I can see at the moment.

Wherever we do, unfortunately spritesheets in the wild tend to be complied irregularly, and often distributed without an atlas. We *could* make an atlas builder tool that lets people divide up an already complied spritesheet, but I don't think we'd want to support that in the long term. 

Probably the best thing to do for irregular spritesheets without an atlas is to type out an atlas. But I understand that's not ideal for a lot of users.

Open to additional ideas. But I'll go ahead and add row/cols like above unless there's an unforeseen problem.

andretchen0:
> asSprite as default

I'd be ok with that. 

I defaulted to `false` because sprites are a bit hard to use in relation to other 3D elements in the scene. They always face the camera.

The current implementation of `<AnimatedSprite />` also makes working with sprites a bit wonky if:

* the frames are different sizes
* the camera is not looking straight down the z
* and the anchor is not the default

Maybe this can be worked around though.

JaimeTorrealba:
Exactly we don't want to support another library (still if there is another free good option, we can send the user there, having a warning that is not under our scope). 
I don't pretend to add full support to non-atlas, as you say they tend to be irregulars and could be a big headhache for us. My intention here is just the control with one row (this pages come with a lot of sprites, and I normally use just one row: https://www.spriters-resource.com/)
If the people want something more complex, they need to use atlases

Regarded asSprite as default I would expect to behave like sprite materials (but I could be wrong, maybe let's ask to someone else)

docs/guide/abstractions/animated-sprite.md
andretchen0: Gotcha. Thanks for the feedback. It's a good catch. I'll clear that up.

andretchen0:
> My intention here is just the control with one row (this pages come with a lot of sprites, and I normally use just one row: https://www.spriters-resource.com/) If the people want something more complex, they need to use atlases

Gotcha. 

Here's the only way I know that will work for spriters resource rips:

* split the sheets into individual images in e.g., GIMP
* stack the images in layers so they line up for animation
* export layers as individual images
* compile individual images to a spritesheet/atlas

At least, that's the only solution that I see. 

----

The rest of this is just details into why that's the case. Feel free to skip. I'm just including it because I don't want you to feel like I'm dismissing your use case.

I've banged my head against the wall with spritesheets from spriters resource before. I think they're mostly compiled by hand from screenshots or other means. As far as I can tell, their purpose is to distribute art to fans, not to make it easy to animate – because they don't include information for animation, i.e., how individual frames line up.

Normally, some care has to be taken to get the animation frames to line up. E.g., here's Wesnoth's documentation for sprite creators.
 
* Units are centered horizontally
* Unit's feet are positioned around 55 pixel down from the top edge of the canvas, lower for taller units if necessary

https://wiki.wesnoth.org/Creating_Unit_Art#Basic_unit_image_specifications

That gets lost in spriters resource rips. Without that information, there's no way to automate the animation, as far as I know. And because spriters resource sprites are cropped for display, the anchor likely changes for every frame – which we don't support (currently) and Drei doesn't support.

I'd love to be wrong though – it'd be great for our users. So if you see a way to use spriters resource spritesheets without recompiling the spritesheet/atlas, don't hesitate to tell me. That would be a welcome thing for me to learn!

docs/guide/abstractions/animated-sprite.md
andretchen0: Is it clearer now? 

(Fwiw, the range is supposed to be given as the `animation` prop. Just in case that's still not clear.)

andretchen0:
@alvarosabu @JaimeTorrealba 

As discussed on Discord, the `asSprite` prop has been removed for now. Same for `<TresSprite />` in the implementation. I'll work them back in once things are squared away with events.

andretchen0:
@JaimeTorrealba 

FYI, for the docs/playground demos, I'm currently using images from my fork of Tres/assets. I needed to modify the atlas somewhat to make the demos clearer, but I didn't want to keep bothering you to approve pushes to the main branch of Tres/assets. 

Once `<AnimatedSprite />` is approved, I'll submit a PR for the final images and atlas.

playground/src/pages/abstractions/AnimatedSprite.vue
```diff
@@ -0,0 +1,230 @@
...
+  definitions: { label: 'Definitions', value: '{}', options: ['{}', '{"idle":"0(10),1-5"}'] },
+  flipX: false,
+  loop: true,
+  paused: false,
+  reversed: false,
+  resetOnEnd: false,
+  centerX: { value: 0.5, min: 0, max: 1, step: 0.01 },
+  centerY: { value: 0.5, min: 0, max: 1, step: 0.01 },
+  scale: { value: 1, min: 0.1, max: 4, step: 0.01 },
+  rotation: { value: [0, 0, 0] },
```
JaimeTorrealba, line 35: Non blocking comment: 
  IDK if leches has steps, but maybe is a good Idea to put lower steps or change to deg (using something like [radToDeg ](https://threejs.org/docs/index.html?q=math#api/en/math/MathUtils.radToDeg)

src/core/abstractions/AnimatedSprite/Atlas.ts
```diff
@@ -0,0 +1,302 @@
...
+  }
+  return result
+}
+
+export function setAtlasDefinitions(atlas: Atlas, definitions: Record<string, string> = {}) {
+  const animations = groupAtlasFramesByKey(atlas.frames)
+  for (const [animationName, definitionStr] of Object.entries(definitions)) {
+    const frames: AtlasFrame[] = getAtlasFrames(atlas, animationName, false)
+    const expanded = expand(definitionStr)
+    for (const i of expanded) {
```
JaimeTorrealba, line 211: Non-blocking:
This component has lots of mental load, maybe is a good idea to put index instead of just i
I have noticed this pattern in other for loops, not only with I also with n, k, v, d or even _
it works, but maybe we can reduce some mental load for next developers :)

src/core/abstractions/AnimatedSprite/AtlasAnimationDefinitionParser.ts
```diff
@@ -0,0 +1,245 @@
...
+          definitionStr,
+          startI,
+        )
+      }
+    }
+    else if (transition === 'START_FRAME_OUT') {
+      if (name === 'COMMA') {
+        transition = 'START_FRAME_IN'
+      }
+      else if (name === 'HYPHEN') {
```
JaimeTorrealba, line 83: Non blocking: 
kind of the same here with the mental load. I would use 
if (name === 'COMMA') transition = 'START_FRAME_IN'
 else if (name === 'HYPHEN') transition = 'END_FRAME_IN'
 
 And reduce some lines. (maybe I'm wrong with this one, but I feel cleaner

src/core/abstractions/AnimatedSprite/AtlasAnimationDefinitionParser.ts
```diff
@@ -0,0 +1,245 @@
...
+        && result[result.length - 1].name === 'NUMBER'
+      ) {
+        result[result.length - 1].value *= 10
+        result[result.length - 1].value += parseInt(c)
+      }
+      else {
+        result.push({ name: 'NUMBER', value: parseInt(c), startI: ii })
+      }
+    }
+    else if (c === ' ') {
```
JaimeTorrealba, line 196: Is there a state for space?

src/core/abstractions/AnimatedSprite/component.vue
```diff
@@ -0,0 +1,223 @@
...
+        <TresMeshBasicMaterial
+          :toneMapped="false"
+          :side="DoubleSide"
+          :map="texture"
+          :transparent="true"
+          :alphaTest="props.alphaTest"
+          :depthWrite="props.depthWrite"
+          :depthTest="props.depthTest"
+        />
+      </TresMesh>
```
JaimeTorrealba, line 219: At the end, you delete the whole implementation of TresSprite?
I thought you'll be commenting it. 
Well no worries, I'm sure you know well the implementation in the future

JaimeTorrealba (approved):
Great job, I made some comments, all non-blocking

1. Why is this PR in draft? What else is missing?

src/core/abstractions/AnimatedSprite/component.vue
andretchen0: Hey @JaimeTorrealba , thanks for having a look! 

> I thought you'll be commenting it.

I try not to leave dead code as comments.

It often gets forgotten, and then other people don't know if someone's coming back to use the code. So it just stays there.

I have the old implementation in my commit history anyway.

src/core/abstractions/AnimatedSprite/AtlasAnimationDefinitionParser.ts
andretchen0: No, it's accepted but doesn't add any meaning, so it gets omitted.

It let users do ...

```
1, 1, 2, 3(1), 2
```

... instead forcing them to write ...

```
1,1,2,3(1),2
```

src/core/abstractions/AnimatedSprite/AtlasAnimationDefinitionParser.ts
andretchen0: > And reduce some lines. (maybe I'm wrong with this one, but I feel cleaner

I'm old school here. Lol.

I also used to prefer ...

```
if (condition) fn()
```

But I've been caught by bugs when adding a line like this ...

```
if (condition) fn()
```

```
if (condition)
  fn()
  fn2()
```

That syntax is legal and doesn't get flagged in my editor. But `fn2()` always runs, regardless of the condition.

That can be a hard bug to track down. And avoided by adding braces. So I add the braces.

JaimeTorrealba: Yep similar to the arrow function, for multi-line you need the brackets, but if is just one line there is no problem
I saw you have a lot with one line, in any case there is just my perception so, if you don't feel comfortable don't worry

andretchen0:
@JaimeTorrealba 

>     1. Why is this PR in draft? What else is missing?

Is it showing up as a draft for you? I see it as an open PR:

Screenshot 2024-02-04 at 20 08 42

andretchen0:
@alvarosabu @JaimeTorrealba 

Do we want to merge in v3 or wait for v4? 

There are no conflicts with v3, but if we wait for v4, `<TresSprite />` should be clickable so we can add that back in.

JaimeTorrealba:
I thought this one was for the v4 (but depend on you, my friend)

Alvaro is coming the 19th. We have time since v4 comes in the second Q approx.

If you decide to merge for the v4 remember to point to the v4 branch and add the tag :)

andretchen0:
> I thought this one was for the v4 (but depend on you, my friend)
> 
> Alvaro is coming the 19th. We have time since v4 comes in the second Q approx.
> 
> If you decide to merge for the v4 remember to point to the v4 branch and add the tag :)

Thanks for the pointers! 😀 And thanks for the PR approval.

I'll wait for v4 then. That'll give us one more thing to announce.

JaimeTorrealba:
Hello @andretchen0, please remember to change the target branch to [v4](https://github.com/Tresjs/cientos/tree/v4).
We can merge it now

andretchen0:
Hey @JaimeTorrealba ,

I updated `AnimatedSprite` for `v4`. Could you please have a look?

Changes to pay attention to are in the recent commit titles. But namely:

* removed the `<Suspense />` from the component, it's up to users to add it
* added `:as-sprite` prop, default is `true`
* removed the @click from the component – it should be handled by the new event system, but will need to check against new event system when it's ready.

Thanks!

alvarosabu:
Hey there @andretchen0 @JaimeTorrealba is this ready to merge?

andretchen0:
@alvarosabu 

> Hey there @andretchen0 @JaimeTorrealba is this ready to merge?

Merged!

PR #337, opened 2024-01-26
feat(app): Add shadow camera helpers to v-light-helper

JaimeTorrealba:


src/core/directives/vLightHelper.ts
```diff
@@ -1,36 +1,66 @@
...
   },
 }
 let currentHelper: any
 let currentInstance: any
+let currentShadowCameraHelper: any
+
+const getHelperIntance = (el: any) =>
+  el.parent.children.find((child: any) => child instanceof currentHelper)
+const getShadowCameraHelperIntance = (el: any) =>
+  el.parent.children.find((child: any) => child instanceof CameraHelper)
```
andretchen0, line 62: I think this approach is causing some problems. If you `el.parent.children.find` there might be multiple instances of a particular `instanceof`, if there are sibling lights.

## Suggestion

LightHelpers and CameraHelpers store a reference to their targets. You could match on those.

src/core/directives/vLightHelper.ts
```diff
@@ -1,36 +1,66 @@
...
+    // if has ShadowCamera
+    if (!currentShadowCameraHelper) return
+    currentShadowCameraHelper = getShadowCameraHelperIntance(el)
+    currentShadowCameraHelper.dispose()
+    el.parent.remove(currentShadowCameraHelper)
   },
 }
 let currentHelper: any
 let currentInstance: any
+let currentShadowCameraHelper: any
```
andretchen0, line 57: I think these variables are out of sync when the code reaches line 29. Iirc:

* Create SpotLight 1, add helper
* Create PointLight 2, add helper
* updated() on Light 1 <-- `getHelperInstance` will return the PointLight helper instance

## Suggestion

If you want to store references, maybe something like this?

```
type LightHelper = DirectionalLightHelper | PointLightHelper | SpotLightHelper | HemisphereLightHelper | RectAreaLightHelper
const lightUuidToShadowCameraHelper: Record<string, CameraHelper> = {}
const lightUuidToLightHelper: Record<string, LightHelper> = {}
```

andretchen0 (changes requested):
Hey @JaimeTorrealba !

I've found some bugs and made a few suggestions for getting rid of them.

There are also a few changes/suggestions for typos, etc.

😎

JaimeTorrealba:
@andretchen0 thanks for taking the time, I like your feedback. I'm going to put this in draw because we're moving the directive section to the core pkg. So I don't know yet what will happen with this one

JaimeTorrealba:
I'll close this one, since now the directives are in the core (I'll re-open there)

PR #339, opened 2024-01-28
feat(app)!: 227 Change the keyboardcontrols implementation

### LINKED ISSUE #227: Improve KeyboardControls
### Description

KeyboardControls could be improved.

Currently, is not supporting [magic keys](https://vueuse.org/core/useMagicKeys/#usemagickeys). And remove the 2D mode
Could use the camera as a group (like in the MouseParallax) for the jump function, and the headBobbing, and avoid fighting with cameras
Improve the usability for thirdCamera Games

### Suggested solution

.

### Alternative

Any other alternative is considered

We don't take mobile functionality here, that will be in another issue

### Additional context

_No response_

### Validations

JaimeTorrealba:
Please check the playground, and play with it, lt me know what you think.

closes #227

### DIFF
--- a/docs/.vitepress/theme/components/KeyboardControlsDemo.vue
+++ b/docs/.vitepress/theme/components/KeyboardControlsDemo.vue
@@ -1,0 +1,13 @@
+<script setup lang="ts">
+import { TresCanvas } from '@tresjs/core'
+import { KeyboardControls, Box } from '@tresjs/cientos'
+</script>
+
+<template>
+  <TresCanvas>
+    <TresPerspectiveCamera />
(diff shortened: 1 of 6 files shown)


andretchen0:
Hey @JaimeTorrealba !

It works well! 

I think these would be welcome features:

* arrow key support
* gradually ramp up to top speed for movement. Maybe implemented using constants like [the Three.js example](https://threejs.org/examples/?q=Controls#misc_controls_pointerlock). Or maybe implemented as reactive props: acceleration, friction, maxSpeed.

Optionally: [head bob!](https://forum.zdoom.org/viewtopic.php?t=48194)

PR #354, opened 2024-02-20
fix: line2 cannot dynamically add the count of points

Steve245270533:
In the original version, the change of line2 can only be based on the number of points during initialization. If you want to add the number of points later, it cannot be rendered correctly.

```vue
<template>
	<Line2
		:points="points"
		:line-width="20"
		color="#82dbc5"
	/>
</template>

<script setup lang="ts">
import {Line2} from '@tresjs/cientos'
import {shallowRef} from "vue";
import { useTresContext } from '@tresjs/core'
import {Vector2} from "three";

const {raycaster, scene} = useTresContext();

const points = shallowRef([[0, 0, 0], [1, 1, 1]])

const mousePoint = new Vector2();
document.body.addEventListener("click", e => {
	mousePoint.x = (e.clientX / window.innerWidth) * 2 - 1;
	mousePoint.y = -((e.clientY / window.innerHeight) * 2 - 1);
	raycaster.value.setFromCamera(mousePoint, raycaster.value.camera);
	const intersects = raycaster.value.intersectObjects(scene.value.children, true);

	if (intersects.length > 0) {
		const {x, y, z} = intersects[0].point;
		points.value = [...points.value, [x, y, z]];
	}
})
</script>
```
eg

### DIFF
--- a/src/core/shapes/Line2.vue
+++ b/src/core/shapes/Line2.vue
@@ -66,7 +66,7 @@ function getInterpolatedVertexColors(vertexColors: VertexColors | null, numPoint
   return iColors
 }
 
-function updateLineGeometry(geometry: LineGeometry, points: Points, vertexColors: VertexColors | null) {
+function updateLineGeometry(geometry: LineGeometry & { _maxInstanceCount?: number }, points: Points, vertexColors: VertexColors | null) {
   const pValues = points.map((p) => {
     if (p instanceof Vector3) {
       return [p.x, p.y, p.z]
@@ -81,7 +81,9 @@ function updateLineGeometry(geometry: LineGeometry, points: Points, vertexColors
       return p
     }
   }).flat()
-  geometry.setPositions(pValues.flat())
+  const flatPositions = pValues.flat()
+  geometry.setPositions(flatPositions)
+  geometry._maxInstanceCount = flatPositions.length / 3
 
   const colors = getInterpolatedVertexColors(vertexColors, points.length).map(c => c.toArray()).flat()
   geometry.setColors(colors)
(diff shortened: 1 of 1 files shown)


andretchen0:
Hi @Steve245270533 

Thanks for the PR! 

I'm getting errors when I run the Line2 from the PR on the playground CatmullRom3, but I'd like to investigate further. 

Could you please file a bug report with a StackBlitz that shows the bug and @ me?

Thanks for the help!

andretchen0:
Hi @Steve245270533 

Thanks for the PR. 

This appears to be a valid bug.

I'm getting some WebGL warnings when running the PR's Line2 on the CatmullRom3 playground example. I'll look into it.

Steve245270533:
> Hi @Steve245270533
> 
> Thanks for the PR.
> 
> This appears to be a valid bug.
> 
> I'm getting some WebGL warnings when running the PR's Line2 on the CatmullRom3 playground example. I'll look into it.

However, I believe both `Lin2.vue` and `CatmullRomCurve3.vue` need to validate the format of the points parameter passed in and provide clear and specific prompts; otherwise, errors will occur when instantiating the `LineGeometry` constructor. For example:
```vue
<Line2
    :points="[]"
    :line-width="20"
    color="#82dbc5"
/>
<CatmullRomCurve3
    :points="[]"
    :segments="40"
    :line-width="10"
    color="#fbb03b"
/>
```
image
image

andretchen0:
Closing. Will not fix Three's limitations.

Hey @Steve245270533 

Thanks again for the PR.

[Three.js doesn't allow buffer attributes to be resized after creation.](https://discourse.threejs.org/t/how-costly-is-allocating-new-bufferattribute-array/2837) On the Discord, it was decided to let Three be Three and not try to hide its (and WebGL's) limitations.

In the future (or if you think this was closed in error) it would be most helpful if you'd file a bug report with a working StackBlitz showcasing the bug, rather than only filing a PR. Thanks in advance.

PR #356, opened 2024-02-22
fix: line2 cannot dynamically add the count of points

Steve245270533:
Hi @andretchen0 , what I mean is that if there is a scene that requires dynamically appending track data, then it cannot be displayed according to the ArrayBuffer size defined during initialization, as shown in the case provided below:
[stackblitz](https://stackblitz.com/edit/vitejs-vite-d8vua4?file=src%2Fcomponents%2FLineTrack.vue)
You can compare the differences between this solution and the official version
In this case, this may be a hack method, but after I personally verified it, it does work.

### DIFF
--- a/src/core/shapes/Line2.vue
+++ b/src/core/shapes/Line2.vue
@@ -66,7 +66,7 @@ function getInterpolatedVertexColors(vertexColors: VertexColors | null, numPoint
   return iColors
 }
 
-function updateLineGeometry(geometry: LineGeometry, points: Points, vertexColors: VertexColors | null) {
+function updateLineGeometry(geometry: LineGeometry & { _maxInstanceCount?: number }, points: Points, vertexColors: VertexColors | null) {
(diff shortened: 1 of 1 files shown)


Steve245270533:
I've worked on a similar project before where I needed to display real-time trajectory lines for objects. Therefore, this data wasn't fixed from the beginning; it kept getting appended over time, resulting in longer and longer line segments. 
So handling this kind of situation doesn't seem to be ideal in the original solution.

andretchen0:
Hey @Steve245270533 

I talked with the team briefly about this the last time it was brought up and the consensus was that we shouldn't patch here, because it was seen as a design decision in THREE, which doesn't allow buffers to be resized.

However, you can add/remove points by forcing a rerender using the `:key` prop. Here's an example:

https://stackblitz.com/edit/tresjs-basic-vwqj9p?file=src%2Fcomponents%2FTheExperience.vue

PR #377, opened 2024-04-06
fix(MeshReflectionMaterial)!: add features/docs, reorganize

### LINKED ISSUE #371: Reflector Material
### Description

As a developer using TresJS, it'd be nice to be able to easily add reflections. Especially for floors, reflections show off user models nicely.

Three's and Cientos' `Reflector` goes part way, but doesn't have a few features, like optional blurring.

### Suggested solution

Equivalent of Drei/R3F's MeshReflectorMaterial

### Alternative

_No response_

### Additional context

_No response_

### Validations

andretchen0:
## Problem

Some of `MeshReflectionMaterial`'s features – e.g., features related to "blur" – weren't working.

## Solution

Rewrite `MeshReflectionMaterial`

Playground at `/materials/reflection-material`

----

## Problem

`MeshReflectionMaterial` was undocumented.

## Solution

Document `MeshReflectionMaterial` at `/guide/materials/mesh-reflection-material.html`

----

## Problem

`blurPass` didn't have a way of disposing its resources, leading to memory leaks.

## Solution

Add dispose to `blurPass` and dispose it when necessary.

----

## Breaking changes

The issue of breaking changes was brought up on the Discord. It was decided that since the material was undocumented up to this point, users probably don't rely on the current implementation, and there was little benefit in avoiding breaking changes.

* `blurPass` and `convolutionMaterial` move inside `MeshReflectionMaterial` and their implementations change. These files were/are undocumented. Their implementations are relatively tightly bound to `MeshReflectionMaterial`'s implementation, so they're not generally useful as written.
* The PR activates a lot of previously non-working features of `MeshReflectionMaterial`. The visual output of the material is much different, for the same props.
* I added some extra props, to hopefully make it easier to get the desired effects. For example, now the user can control how the strength of the blur effect on rough and smooth surfaces, separately. But that meant changing prop names.

---

## Related issues

Some `NOTE:`/`TODO:`s have been left in the files concerning issues filed against the Tres core.

In particular, [the Tres core currently doesn't reattach recompiled materials](https://github.com/Tresjs/tres/issues/615). Some props changes have to trigger a recompile of the material in order to take effect. Some of those are due to `THREE.MeshStandardMaterial`'s implementation, so can't be worked around here, while still retaining the desired functionality/reactivity.

The material logs warnings if users change props in ways that trigger a recompile *specific to this material* – e.g., toggling the blur or "depth".

---

## Room for improvement?

The material extends `MeshStandardMaterial`.

The material ran into reactivity/recompile issues when "inherited" props/defaults were not written out in `MeshStandardMaterial`.

---

closes #371


JaimeTorrealba (approved):
Question as this one have breaking changes, it would be merged into v4? or no problem

andretchen0:
> Question as this one have breaking changes, it would be merged into v4? or no problem

Hey @JaimeTorrealba , I'm merging into `v4` because of the breaking changes.

Do you see a problem with that?

JaimeTorrealba:
Not at all, is just I didn't see the label v4,

andretchen0:
> Not at all, is just I didn't see the label v4,

Ah, gotcha. I changed the "merge into" branch in Github, but didn't add the label.

Sorry for the confusion. I just added the `v4` label.

Issue #384, opened 2024-04-18
CameraControls component doesn't support mouseButtons and touches props


whereiswolf:
### Describe the bug

[`CameraControls`](https://cientos.tresjs.org/guide/controls/camera-controls.html) component ignores `mouseButtons` and `touches` properties.

### Reproduction

[...]

PR #385, opened 2024-04-18
fix: camera controls ignored props

whereiswolf:
- Fixes https://github.com/Tresjs/cientos/issues/384
- Adds support for `mouseButtons` and `touches` prop.
- Adds default `mouseButtons` and `touches` prop values based on https://yomotsu.github.io/camera-controls/examples/config.html
- Adds support for partial overrides of `mouseButtons` and `touches`.
- Updates documentation of `CameraControls`
---
Feel free to point out at any mistakes and discrepancies from conventions. I have a strong React background, and some aspects of Vue are kinda new to me 😜

### DIFF
--- a/src/core/controls/CameraControls.vue
+++ b/src/core/controls/CameraControls.vue
@@ -4,7 +4,9 @@ import { ref, watchEffect, onUnmounted, toRefs, computed } from 'vue'
 import type {
   PerspectiveCamera,
   OrthographicCamera,
-  Object3D } from 'three'
+  Object3D,
+  Camera,
+} from 'three'
 import {
   Box3,
   MathUtils,
@@ -19,6 +21,7 @@ import {
 } from 'three'
 import { useRenderLoop, useTresContext } from '@tresjs/core'
 import { useEventListener } from '@vueuse/core'
+import { isPerspectiveCamera, isOrthographicCamera } from '../../utils/types'
 
 export interface CameraControlsProps {
   /**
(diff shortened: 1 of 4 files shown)


JaimeTorrealba (changes requested):
Hey this work perfect, is very professional, thanks for this one.

But remember to update the docs for the component (adding the new props and maybe and example of the type that you can use, better be really explicit for the final user).

JaimeTorrealba:
I agree with you, I would do it @whereiswolf (maybe you can export it in the defaultExpose ¿?)

whereiswolf:
@JaimeTorrealba quick summary of the recent changes:

- Updated `CameraControls` docs with "User input config" section.
- Added logic to check the camera type, to set the default values in the same way as `camera-controls` does.
- Added `CameraControlsClass` re-export of `camera-controls` (if you have a better name suggestion please lmk 😄 )

If anything's unclear or should be changed, I'm happy to work on that 😉

src/core/controls/CameraControls.vue
```diff
@@ -411,6 +433,43 @@ defineExpose({
...
+      ? CameraControls.ACTION.TOUCH_DOLLY_TRUCK
+      : isOrthographicCamera(camera)
+        ? CameraControls.ACTION.TOUCH_ZOOM_TRUCK
+        : CameraControls.ACTION.NONE
+  ),
+  three: CameraControls.ACTION.TOUCH_TRUCK,
+  ...touches,
+})
+
+export { default as CameraControlsClass } from 'camera-controls'
```
JaimeTorrealba, line 470: Sorry for the late I was off this weekend.

I love the result. The only thing I am not sure is the name "CameraControlsClass" could that create confusion for some users? (since ThreeJs is basically made in classes) 

But I don't have a better name, maybe CameraControlsPointers?

whereiswolf: Yeah, I wasn't sure about that either, what about `ThreeCameraControls` or `BaseCameraControls`? Maybe I miss something, but `Pointers` adds more ambiguity in my opinion 😛.

As one said:
> There are only two hard things in Computer Science: cache invalidation and naming things.

JaimeTorrealba: I like BaseCameraControls :)

whereiswolf: Updated 😉

JaimeTorrealba (approved):
Great work man!

PR #424, opened 2024-05-16
fix: vue attrs hyphenation issues

alvarosabu:


andretchen0 (approved):
The Leches values weren't working in `SkyDemo`. I updated them:

`.value.value` -> `.value`

PR #444, opened 2024-07-10
feat(app): Add the option for x and y in mouseparallax component

JaimeTorrealba:
Small PR for improving the mouse-parallax component

v4 branch is selected


docs/guide/abstractions/mouse-parallax.md
```diff
@@ -10,9 +10,11 @@
 
 You only need to import it and add it to your template as `<MouseParallax />`. Additionally, you can pass the following props:
 
-`factor` is a number to increase the movement range of the camera. `ease` is a number that smoothes the movement. You can also disable the effect with the `disabled` prop.
+`factor` is a number to increase the movement range of the camera. This could be an array of two values corresponding to the x and y values, in that order, ej. `:factor=[x,y]` .
```
andretchen0, line 13: Wording: 

I'm guessing "ej." is for "ejemplo", right?

I think here we'd write, "in other words" or "that is to say" or "i.e.". Or maybe just leave it out altogether:

```
This could be an array of two values corresponding to the x and y values, in that order: `:factor=[x,y]`.
```

docs/guide/abstractions/mouse-parallax.md
```diff
@@ -10,9 +10,11 @@
...
 
-`local` is a boolean that enables movement based on the position of the mouse on the element rather than the window.
+`ease` is a number that smooths the movement. This could be an array of two values corresponding to the x and y values, in that order, ej. `:ease=[x,y]` .
```
andretchen0, line 15: Wording:

Same as above. Maybe omit the "ej.":

```
This could be an array of two values corresponding to the x and y values, in that order: `:ease=[x,y]`.
```

playground/src/pages/abstractions/MouseParallaxDemo.vue
```diff
@@ -24,10 +24,10 @@ const gl = {
     <TorusKnot>
       <TresMeshNormalMaterial />
     </TorusKnot>
-    <MouseParallax :factor="3" />
+    <MouseParallax :factor="[30, 15]" :ease="[3, 0.1]" />
     <TresAmbientLight :intensity="1" />
   </TresCanvas>
-  <TresCanvas v-bind="gl">
+  <!-- <TresCanvas v-bind="gl">
```
andretchen0, line 30: We sometimes have issues in Tres caused by comments.

Since handling comments isn't what this playground is showing, can we remove this comment?

src/core/abstractions/MouseParallax.vue
```diff
@@ -21,15 +21,15 @@ export interface MouseParallaxProps {
    * @memberof MouseParallaxProps
    *
    */
-  factor?: number
+  factor?: number | [number, number]
```
andretchen0, line 24: The JSDoc type `@type {number}` doesn't agree with the TS type `number | [number, number]`.

VS Code displays both types, so I think we should just remove the JSDoc type to make future updates easier.

src/core/abstractions/MouseParallax.vue
```diff
@@ -21,15 +21,15 @@ export interface MouseParallaxProps {
...
   /**
    * The factor to smooth the mouse movement by.
    * @type {number}
    * @default 2.5
    * @memberof MouseParallaxProps
    *
    */
-  ease?: number
+  ease?: number | [number, number]
```
andretchen0, line 32: Same as above. 

The JSDoc type `@type {number}` doesn't agree with the TS type `number | [number, number]`.

VS Code displays both types, so I think we should just remove the JSDoc type to make future updates easier.

src/core/abstractions/MouseParallax.vue
```diff
@@ -65,12 +65,15 @@ const { width, height } = local.value
 
 const cameraGroupRef = ref<Group>()
 
-const cursorX = computed(() => (x.value / width.value - 0.5) * factor.value)
-const cursorY = computed(() => -(y.value / height.value - 0.5) * factor.value)
+const _factor = Array.isArray(factor.value) ? factor.value : [factor.value, factor.value]
+const _ease = Array.isArray(ease.value) ? ease.value : [ease.value, ease.value]
```
andretchen0, line 69: It seems that the props `factor` and `ease` aren't reactive.

`_factor` and `_ease` will need a `watch` or some other way to detect prop changes. Or you could add a ternary here:

```
const cursorX = computed(() => (x.value / width.value - 0.5) * _factor[0])
```

Here's a demo showing that. This should alternate between zooming around with a value of `[10, 10]` and stopping with `0`.

```vue
<script setup lang="ts">
import { TresCanvas } from '@tresjs/core'
import { MouseParallax, TorusKnot } from '@tresjs/cientos'

const factorOrFactors = shallowRef<number | [number, number]>([10, 10])
setInterval(() => {
  factorOrFactors.value = factorOrFactors.value === 0 ? [10, 10] : 0
},
  1000
)
</script>

<template>
  <TresCanvas>
    <TresPerspectiveCamera :position="[0, 0, 7.5]" />
    <TorusKnot>
      <TresMeshNormalMaterial />
    </TorusKnot>
    <MouseParallax :factor="factorOrFactors" :ease="3" />
  </TresCanvas>
</template>
```

andretchen0 (changes requested):
Hey @JaimeTorrealba . Welcome back! This is a useful change!

src/core/abstractions/MouseParallax.vue
JaimeTorrealba: Good one, yes I've destroy the reactivity with my last update, thanks for this one

src/core/abstractions/MouseParallax.vue
JaimeTorrealba: Do you think we should remove? I mean JSDocs accept multiples types adding just a pipe operator.
https://jsdoc.app/tags-type

Don't see why remove it

andretchen0: Hey @JaimeTorrealba ,

Keeping it means we have 2 "sources of truth" for the type. But sure, as long as they match, then it's ok to keep both.

src/core/abstractions/MouseParallax.vue
```diff
@@ -64,13 +64,24 @@ const { width, height } = local.value
   : useWindowSize()
 
 const cameraGroupRef = ref<Group>()
```
andretchen0, line 66: (Non-blocking)

Sorry, I missed this earlier. This should probably be a `shallowRef`.

> shallowRef() is typically used for performance optimizations of large data structures, or integration with external state management systems.

https://vuejs.org/api/reactivity-advanced.html#shallowref

Since all Tres elements have a link to `context` – which includes `Scene` and all its descendants – it probably qualifies as a "large data structure."

andretchen0 (approved):
Looks good!

I caught one thing that wasn't changed in this PR, but since we're looking at component, we might as well change it. 

I won't block the PR though.

PR #449, opened 2024-07-17
fix: reimplement ContactShadows for v4

andretchen0:
## Problem

`ContactShadows` is not working with Tres `v4`.

## Solution

Reimplement `ContactShadows` from the original THREE.js source.


alvarosabu (approved):
This is great @andretchen0 so clean. I prefer this approach 1000x more than the previous one. Thanks for refactor it

PR #452, opened 2024-07-22
chore: remove on demand composable

alvarosabu:


src/core/controls/KeyboardControls.vue
andretchen0, line 1: There's an `invalidateOnDemand()` on 114.

src/core/staging/Fit.vue
```diff
@@ -20,12 +20,17 @@ export interface Props {
...
+  {
+    into: () => new Box3(new Vector3(-0.5, -0.5, -0.5), new Vector3(0.5, 0.5, 0.5)),
+    precise: false,
+  },
+)
+
+const { invalidate } = useTresContext()
 
-useOnDemandInvalidation(props)
+watch(props, () => invalidate())
```
andretchen0, line 33: ## Problem

This `watch` makes the demo drop frames when `into` is an `Object3D`.

* `pnpm run playground`
* go to http://localhost:5173/staging/fit
* click last radio button – label starts with`Object`

The demo doesn't drop frames on my machine if the `watch` is removed.

## Problem

The component exports `fit` and `update` functions that can be called by the user. Those functions do not `invalidate()`.

## Solution

* Remove the `watch`.
* Put the `invalidate()` in `fit()`. (line 41)

andretchen0 (changes requested):

src/core/controls/MapControls.vue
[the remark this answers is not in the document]
alvarosabu: I agree we can do this in the scope of another PR, seems to be a problem overall in the controls

andretchen0 (approved):

PR #459, opened 2024-07-31
fix: typescript issues

alvarosabu:


src/core/controls/CameraControls.vue
```diff
@@ -407,7 +408,7 @@ extend({ CameraControls })
 watchEffect(() => {
   addEventListeners()
   if (controlsRef.value && makeDefault.value) {
-    controls.value = controlsRef.value
+    controls.value = controlsRef.value as unknown as EventDispatcher<object> & { enabled: boolean }
```
alvarosabu, line 411: @andretchen0 I think for this one it may be worth creating a proper type and export it from the core here

https://github.com/Tresjs/tres/blob/e6389463c81a58aca1769bf1d83c53c2bf9f512e/src/composables/useTresContextProvider/index.ts#L57

If you agree I will create a PR on the core for this

andretchen0 (approved):

andretchen0:
I had 2 questions and posted them to the Discord.

But otherwise, it appears this is ready to go.

PR #460, opened 2024-08-01
chore(linter): add TypeScript error check to package.json and pre-push hook

andretchen0:
## Problem

TypeScript errors are not caught by Cientos' current pre-push hook. This has led to an accumulation of TypeScript errors in the codebase over time.

## Solution

* Add `pnpm lint:ts` to print out a list of TypeScript errors
* Move current `pnpm lint` to `pnpm lint:eslint`
* Modify `pnpm lint` to check for TypeScript errors
* Add TypeScript error check in Husky pre-push hook and – by default – abort push if errors are present.


alvarosabu:
Hey @andretchen0 is this PR still needed after we removed the pre-push strategies? If its let me know and we re-open it

PR #462, opened 2024-08-07
feat: 393 lod

JaimeTorrealba:
close #393 
Hi, I choose a different approach from r3f (because I personally don't like so much the `details` component or how it works in drei). Please feel free to give feedback.

Note: right now the component is not reactive (what do you think about this component been reactive anyway, I believe just for ... is useful) the reason for this is that is a little difficult to handle the remove-create new levels because there's no removeLevel method in LOD class.

I create a new issue in Threejs https://github.com/mrdoob/three.js/issues/29074 to addres this one.

TWO questions.

1. Should I merge this one for v4? Or wait until v4 is merged to main?
2. Should I create a new section of `performarce` in cientos? We have some component that could be moved there (stats, statsGl, bakedShadow, etc.)


docs/guide/misc/lod.md
```diff
@@ -0,0 +1,26 @@
+# LOD (level of detail)
+
+<DocsDemo>
+  <LODDemo />
+</DocsDemo>
+
+**Cientos** provides a component called `LOD` that show meshes with more or less geometry based on distance from the camera. based on: [LOD](https://threejs.org/docs/index.html?q=lod#api/en/objects/LOD)
```
alvarosabu, line 7:
```suggestion
**Cientos** provides a component called `LOD` (Level of Detail) which reduces the resolution of the geometry based on distance from the camera, based on: [LOD](https://threejs.org/docs/index.html?q=lod#api/en/objects/LOD)
```

docs/guide/misc/lod.md
```diff
@@ -0,0 +1,26 @@
...
+
+Every level is associated with a group, and rendering can be switched between them at the distances specified. Typically you would create, say, three meshes, one for far away (low detail), one for mid range (medium detail) and one for close up (high detail).
```
alvarosabu, line 9:
```suggestion
Every level is associated with a group, and rendering can be switched between them at the distances specified. Typically you would create, say, three meshes: 
- one for far away (lesser detail)
- one for mid-range (medium detail) 
- one for close-up (higher detail).
```

alvarosabu: @JaimeTorrealba It's easier to read on a list.

docs/guide/misc/lod.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Basic usage
+
+You call a `<LOD >` component choose a `distance` prop, them you put inside all the component that will be rendered at that distance
```
alvarosabu, line 13:
```suggestion
When using the `<LOD >` component you can pass a `distance` prop, then you put inside all the components that will be rendered at that distance
```

src/core/misc/LOD.vue
```diff
@@ -0,0 +1,58 @@
...
+  hysteresis: 0,
+})
+
+const { seek } = useSeek()
+const { scene } = useTresContext()
+
+const LODInstance = shallowRef()
+const wrapperRef = shallowRef()
+
+LODInstance.value = seek(scene.value, 'type', 'LOD')
```
alvarosabu, line 37: What would happen if I add two LODs with the same `name` ?

alvarosabu (changes requested):
This is exciting, nice addition @JaimeTorrealba

docs/guide/misc/lod.md
JaimeTorrealba: I like it but, I will add decrease or increase ...

src/core/misc/LOD.vue
JaimeTorrealba: Absolute nothing, the name is just for helping you

docs/guide/misc/lod.md
JaimeTorrealba: Here the prop distance is not optional, so I will put "need" instead of "can" if you agree

alvarosabu: Agree

alvarosabu:
Hey @JaimeTorrealba when you have a chance can you check the conflicts, I think we are good to go after that

JaimeTorrealba:
@alvarosabu Please mark as resolved all the comments and give me an approval if you think is ready, I already solved the conflicts.

alvarosabu:
Hi @JaimeTorrealba when you are back can you please check the conflicts and check why the docs and the linter are broken? Thanks

src/core/misc/LOD.vue
andretchen0, line 37: ## Implementation

There's only a single `THREE.LOD` instance here, but that won't work for scenes where some elements are close to the camera and some are far away.

Looking at the THREE example code, there is a `THREE.LOD` instance for each object.

https://github.com/mrdoob/three.js/blob/37d6f280a5cd642e801469bb048f52300d31258e/examples/webgl_lod.html#L71

https://threejs.org/examples/#webgl_lod

src/core/misc/LOD.vue
```diff
@@ -0,0 +1,58 @@
...
+
```
andretchen0, line 38: ## API

It seems to me that there's no way to signify that 2 Cientos `LOD`s should be handled differently. 

Below, I want `Box` and `Sphere` to be handled differently.

Here's what I tried:

```html
<script setup lang="ts">
import { CameraControls, LOD, Sphere } from '@tresjs/cientos'
import { shallowRef, watch } from 'vue'
import { TresCanvas } from '@tresjs/core'
import { NoToneMapping, SRGBColorSpace } from 'three'

const gl = {
  clearColor: '#111',
  outputColorSpace: SRGBColorSpace,
  toneMapping: NoToneMapping,
}

const lodRef = shallowRef()

watch(lodRef, (value) => {
  // eslint-disable-next-line no-console
  console.log('jaime ~ LOD LEVELS ~ value:', value.instance)
})
</script>

<!-- <Sphere name="sphere" :args="[1, 16]" :position="[-1, 0, 0]">
  <TresMeshNormalMaterial wireframe />
</Sphere> -->
<!-- -->
<template>
  <TresCanvas v-bind="gl">
    <TresPerspectiveCamera :position="[0, 2, 5]" />
    <CameraControls :distance="50" />
    <LOD ref="lodRef" :distance="0" name="high">
      <Sphere name="sphere" :args="[1, 16]" :position="[-1, 0, 0]">
        <TresMeshNormalMaterial wireframe />
      </Sphere>
    </LOD>
    <LOD :distance="15" name="medium">
      <Sphere name="sphere" :args="[1, 8]" :position="[-1, 0, 0]">
        <TresMeshNormalMaterial wireframe />
      </Sphere>
    </LOD>

    <LOD :distance="30" name="low">
      <Sphere name="sphere" :args="[1, 4]" :position="[-1, 0, 0]">
        <TresMeshNormalMaterial wireframe />
      </Sphere>
    </LOD>

    <LOD :distance="0" :position="[0, 0, -10]">
      <TresMesh name="box" :position="[2, 0, 0]">
        <TresBoxGeometry :args="[1, 1, 1]" />
        <TresMeshBasicMaterial wireframe :color="0xFF0000" />
      </TresMesh>
    </LOD>
    <LOD :distance="5" :position="[0, 0, -10]">
      <TresMesh name="box" :position="[1, 0, 0]">
        <TresBoxGeometry :args="[1, 1, 1]" />
        <TresMeshBasicMaterial wireframe :color="0xFF0000" />
      </TresMesh>
    </LOD>
  </TresCanvas>
</template>
```

andretchen0 (changes requested):
Hey @JaimeTorrealba ,

I think this is a welcome feature for Cientos.

But I think there need to be some changes to the API and the implementation to allow multiple `LOD`s.

See comments.

src/core/misc/LOD.vue
andretchen0, line 1: ## Reparenting bug

The current implementation reparents objects wrapped with `LOD`.

The 2 visible objects in this scene should move, because they're in a group that moves. But the object in the `LOD` doesn't move.

https://github.com/user-attachments/assets/b44f5f03-9656-4106-9553-bb4dab8ad1f6

Here's the code.

```vue
<script setup lang="ts">
import { CameraControls, LOD, Sphere } from '@tresjs/cientos'
import { shallowRef, watch } from 'vue'
import { TresCanvas } from '@tresjs/core'
import { NoToneMapping, SRGBColorSpace } from 'three'

const gl = {
  clearColor: '#111',
  outputColorSpace: SRGBColorSpace,
  toneMapping: NoToneMapping,
}

const groupRef = shallowRef({ position: { x: 0 } })
setInterval(() => groupRef.value.position.x = Math.cos(Date.now() * 0.001), 100)
</script>

<template>
  <TresCanvas v-bind="gl">
    <TresPerspectiveCamera :position="[0, 2, 5]" />
    <CameraControls :distance="50" />

    <TresGroup ref="groupRef">
      <TresMesh>
        <TresCylinderGeometry :args="[1, 1, 1]" />
        <TresMeshBasicMaterial wireframe :color="0xFF0000" />
      </TresMesh>
      <LOD :distance="100" :position="[0, 0, -10]">
        <TresMesh name="box" :position="[2, 0, 0]">
          <TresBoxGeometry :args="[1, 1, 1]" />
          <TresMeshBasicMaterial wireframe :color="0xFF0000" />
        </TresMesh>
      </LOD>
    </TresGroup>
  </TresCanvas>
</template>
```

andretchen0 (changes requested):

src/core/misc/LOD.vue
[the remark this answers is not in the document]
JaimeTorrealba: Hi thanks for going so in deep with this, my implementation here was different for the R3F, because I don't feel intuitive how they handle it, I think it is very confusing an error-prone to match the LOD for the order. That is why I propose something different :)

src/core/misc/LOD.vue
JaimeTorrealba: We can handle this differently, thanks for reaching this.

src/core/misc/LOD.vue
JaimeTorrealba: Unmm, this behavior is only applied for the LOD? It comes for Tres or for Threejs?

There are other similar components that use <TresGroup /> as wrappers, for example scrollControls have you tried if this behavior is only for Tres and LOD? Or get repeated in other contexts?

andretchen0: Hey @JaimeTorrealba ,

> Unmm, this behavior is only applied for the LOD? It comes for Tres or for Threejs?

Afaik, the reparenting happens because the `<LOD>` implementation only ever creates one `THREE.LOD` instance. `THREE.LOD` is a subclass of `Object3D`. When the `<LOD>` children are added to the single `THREE.LOD`, they're removed from their original parent:

Here's the `add` in `THREE.LOD` code:

https://github.com/mrdoob/three.js/blob/beab9e845f9e5ae11d648f55b24a0e910b56a85a/src/objects/LOD.js#L71

Here's where `Object3D.add(child)` removes the child from its current parent:

https://github.com/mrdoob/three.js/blob/841ca14e89f3ec925e071a321958e49a883343c0/src/core/Object3D.js#L336 

> There are other similar components that use as wrappers

I don't know about the others.

But the behavior here doesn't match what I expected to happen.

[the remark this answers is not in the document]
andretchen0: I think the R3F approach is a good compromise.

Matching `distances` to children like below is a bit of a pain, but I think it's doable:

```html
<Detailed distances={[0, 10, 20]} {...props}>
  <mesh geometry={highDetail} />
  <mesh geometry={mediumDetail} />
  <mesh geometry={lowDetail} />
</Detailed>
```

I think the proposed API has bigger problems.

## Breaks the usual Tres mapping: 1 `<LOD />` maps to 1 `THREE.LOD`

In the THREE source, every `THREE.LOD` is an `Object3D`. For most `Object3D`s, Tres/Cientos maps 1 template tag to 1 `Object3D`. That would look something like the R3F API where an `<LOD>` has children for each "level".

But the proposed API here has multiple `<LOD />` tags for a single `THREE.LOD`. That leads to problems like reparenting.

## Record keeping

We need to be able to have multiple `THREE.LOD`s. Since many `<LOD />`s map to 1 `THREE.LOD`, you'll have to specify that mapping. Maybe like:

```html
    <LOD id="lod-0"></LOD>
    <LOD id="lod-0"></LOD>

    <LOD id="lod-1"></LOD>
    <LOD id="lod-1"></LOD>
```

I find this to be more inconvenient than managing `distances` in the R3F API.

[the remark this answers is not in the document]
JaimeTorrealba: Understood, I'll close this PR for now, I would try some ideas here maybe re-open it in the future.

PR #521, opened 2024-11-08
fix(dts): Fix build dts error (fix #520)

### LINKED ISSUE #520: Miss typedef for CameraControls/MapControls/OrbitControls
### Describe the bug

I am import `OrbitControls` by using TypeScript, but the typedef of `OrbitControls` is any

```ts
import { OrbitControls } from '@tresjs/cientos'

type ComponentInstance = typeof OrbitControls
type ControlsInstance = ComponentInstance['instance']
//   ^? This is any
```

I found the build source doesn't include the file `OrbitControls.vue.d.ts`.

image

### Reproduction

https://stackblitz.com/edit/tresjs-basic-grkvtv?file=package.json,src%2Fmain.ts

### Steps to reproduce

Open `main.ts`, then check `ControlsInstance`'s typedef.

### System Info

```shell
@tresjs/cientos is v4.0.3
```

### Used Package Manager

pnpm

### Code of Conduct

0x-jerry:
Before this PR, build dts for `CameraControls/MapControls/OrbitControls` components will failed because of `error TS7056`, this should fix issue #520.

```
src/core/controls/OrbitControls.vue:797:1 - error TS7056: The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.

797 export default (await import('vue')).defineComponent({
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
798 setup() {
    ~~~~~~~~~
... 
805 props: {} as __VLS_WithDefaults<__VLS_TypePropsToOption<__VLS_PublicProps>, typeof __VLS_withDefaultsArg>,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
806 });
```

After this PR, the error `TS7056` will disappear.


alvarosabu (approved):
Hi @0x-jerry thanks for solving this, I was losing my mind with this issue never imagined that was actually a problem of using a `ref` instead of a `shallowRef` for returning the instance, it make sense, the controls are probably really complex objects with tons of nested properties that was making the the ts compiler cry a little.

OrbitControls is a stateful class that manages its own updates
We don't need reactivity for its internal properties
We're already handling updates through the change, start, and end events
It's more performant since we're not creating unnecessary reactive wrappers

Thanks

PR #523, opened 2024-11-18
fix(Line2): allow points to be added/removed, fix defineExpose ref

andretchen0:
## Problem

After initial creation, additional points could not be added to Cientos' Line2 component. The on-screen component would not display the additional points.

## Solution

Recreate the Line2 geometry to accommodate the additional points.

## Problem

The `defineExpose`d `instance` is `undefined` because here ...

```vue
<template>
  <primitive
    :ref="lineRef"
    :object="line"
  />
</template>
```

... `:ref` should be `ref`.

## Solution

Remove the colon from `:ref`.

## Meta

Closes Tresjs/tres#1142

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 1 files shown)


andretchen0:
Mysterious Netlify errors here. Seemingly unrelated to the PR, according to the Netlify build logs. Closing and resubmitting a new PR.

PR #531, opened 2024-11-22
feat(GradientTexture): add component, demo, docs

andretchen0:
Port of Drei's `<GradientTexture>`. Adds:

* component
* demo
* docs

### DIFF
--- a/src/core/abstractions/GradientTexture.vue
+++ b/src/core/abstractions/GradientTexture.vue
@@ -1,0 +1,113 @@
+<script setup lang="ts">
+import { useTres } from '@tresjs/core'
+import * as THREE from 'three'
+import { isReactive, shallowRef, watch } from 'vue'
+
+// NOTE: Source
+// https://github.com/pmndrs/drei/blob/master/src/core/GradientTexture.tsx
+
+export type GradientType = 'linear' | 'radial'
+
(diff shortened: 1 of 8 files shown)


src/core/abstractions/GradientTexture.vue
```diff
@@ -0,0 +1,113 @@
...
+
+if (isReactive(props.stops)) {
+  watch(props.stops, () => update(canvas))
+}
+
+defineExpose({ instance: textureRef })
+</script>
+
+<template>
+  <TresCanvasTexture ref="textureRef" :color-space="useTres().renderer.value.outputColorSpace" :args="[canvas]" :attach="props.attach" />
```
alvarosabu, line 112: For some reason, the playground doesn't run @andretchen0 

Maybe is the use of `useTres` directly on the template, better done in the script tag

Screenshot 2024-12-04 at 09 16 33

alvarosabu (changes requested):

src/core/abstractions/GradientTexture.vue
andretchen0: Good catch. Thanks. Yes, that was the problem.

alvarosabu (approved):

PR #532, opened 2024-11-23
Outline

andretchen0:


src/core/abstractions/Outline/component.vue
```diff
@@ -0,0 +1,126 @@
...
+  renderOrder: 0,
+  thickness: 0.05,
+  angle: Math.PI,
+})
+
+const groupRef = shallowRef()
+
+defineExpose({ instance: groupRef })
+
+const material = new OutlineMaterialImpl({ ...props })
```
alvarosabu, line 49: Non-blocking.

When deconstructing the props they are goign to lose the reactivity. I guess this is intended since you are watching the props to change the material right?

src/utils/shaderMaterial.ts
```diff
@@ -0,0 +1,82 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+import * as THREE from 'three'
```
alvarosabu, line 25: Better to deconstruct what we need from Three for tree-shaking instead of importing it all.

alvarosabu (changes requested):
Just left one blocking about the `shaderMaterial` to deconstruct three imports

src/utils/shaderMaterial.ts
andretchen0: Sure thing. Done.

src/core/abstractions/Outline/component.vue
andretchen0: Right. The props are `watch`ed and update the material. 

https://github.com/user-attachments/assets/fb7f26f6-36af-4575-888d-ca852f93015b

andretchen0: Updated the playground demo to add reactivity.

PR #536, opened 2024-11-24
feat(ScreenSpace): add component, demo, docs

andretchen0:
Port of Drei's `<ScreenSpace />`. Adds:

* component
* demo
* docs


alvarosabu (approved):
@andretchen0 I just changed a minor thing really quick in the last commit, you don't really need to use props on vue templates.

PR #539, opened 2024-11-26
feat(TextTroika): add component, demo, docs

andretchen0:
Wrapper for [troika-three-text](https://github.com/protectwise/troika/tree/1af247e61d2f9f884728a7c77914bc8255c4e7cb/packages/troika-three-text) as `<TextTroika />`. Adds:

* component
* demo
* docs

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -65,6 +65,7 @@ export default defineConfig({
         text: 'Abstractions',
         items: [
           { text: 'Text3D', link: '/guide/abstractions/text-3d' },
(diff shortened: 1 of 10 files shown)


alvarosabu:
Hi, @andretchen0 I'm not convinced we should include this external package since it adds ~166 KB of code. Cientos bundle size is almost 1MB now (986.06 kB).

Maybe it´s time to consider having an integrations subpackage where these dependencies are peer so the user needs to install them along `cientos` to work and they are not included in the bundle

Screenshot 2025-01-06 at 11 07 32


Screenshot 2025-01-06 at 11 06 46

PR #543, opened 2024-11-28
feat(Helper): add component, demo, docs

andretchen0:
Ports Drei's `<Helper />` to Cientos. Adds:

* component
* demo
* docs


src/core/controls/Helper/useHelper.ts
```diff
@@ -0,0 +1,45 @@
...
+type HelperType = Object3D & { update: () => void, dispose: () => void }
+type HelperConstructor = new (...args: any[]) => any
+
+export function useHelper<T extends HelperConstructor>(
+  object3D: MaybeRefOrGetter<Object3D | null | undefined | false>,
+  helperConstructor: T,
+  ...args: any[]
+) {
+  const helper = shallowRef<HelperType>()
+  const scene = useTres().scene
```
alvarosabu, line 18:
```suggestion
  const { scene } = useTres()
```

src/core/controls/Helper/useHelper.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  let currentHelper: HelperType = undefined!
+
+  watchEffect(() => {
+    if (object3D && toValue(object3D) && helperConstructor) {
+      helper.value = currentHelper = new (helperConstructor as any)(toValue(object3D), ...args)
```
alvarosabu, line 24: @andretchen0 any chance we can avoid using `any` assertion here?

alvarosabu (changes requested):

src/core/controls/Helper/useHelper.ts
andretchen0: No problem. I made the change.

For what it's worth, the effect of the change is purely cosmetic. The resulting type of `helper.value` is still

```ts
Ref<HelperType | undefined, HelperType | undefined>.value: HelperType | undefined
```

src/core/controls/Helper/useHelper.ts
andretchen0: Done

PR #544, opened 2024-11-29
feat(Align): add component, demo, docs

andretchen0:
Ports Drei's `<Center />` to Cientos as `<Align />`. Adds:

* component
* demo
* docs


andretchen0:
@alvarosabu 

I called this `<Align>` instead of `<Center>` (like Drei). Thoughts?

For context, it puts all child elements in a `Box3`, then aligns the box to the origin. It doesn't align the children individually, which "align" might suggest.

So, "align" isn't perfect, but I thought maybe it was better than "center".

PR #553, opened 2024-12-03
feat(MarchingCubes): add component, demo, docs

andretchen0:
Port of THREE/Drei's `<MarchingCubes />` to Cientos. Adds:

* component
* demo
* docs


JvanderHeide:
Would love to see this merged

andretchen0:
> Would love to see this merged

Good to hear!

src/core/abstractions/MarchingCubes/MarchingCubes.vue
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const props = withDefaults(defineProps<MarchingCubesProps>(), {
+  resolution: 28,
+  maxPolyCount: 10000,
+  enableUvs: false,
+  enableColors: false,
+})
+
+const marchingCubes = computed(() => new MarchingCubesImpl(props.resolution, null as unknown as Material, props.enableUvs, props.enableColors, props.maxPolyCount))
```
alvarosabu, line 22: mmm... @andretchen0 I don't think we should pass a null and cast it as a material.

Should we use a `new MeshBasicMaterial` as default or maybe enable a prop to pass a material? How does the MarchingCubes get the material from? Like this?

```
<MarchingCubes :enable-colors="true" :resolution="40" :max-poly-count="40000">
      <MarchingPlane v-if="c.hasWallX.value.value" plane-type="x" />
      <MarchingCube
        :position="[c.x.value.value, c.y.value.value, c.z.value.value]"
        :strength="c.strength.value.value"
        :subtract="c.subtract.value.value"
      />
      <MarchingCube ref="r0" color="red" />


      <TresMeshBasicMaterial v-if="c.enableColors.value.value" :vertex-colors="true" />
      <TresMeshPhongMaterial v-else specular="#111111" :shininess="30" color="#049ef4" :reflectivity="1" />
    </MarchingCubes>
```

alvarosabu (changes requested):
I really like this abstraction. I just have some questions about how to use materials and some typescript stuff long the way

src/core/abstractions/MarchingCubes/MarchingCubes.vue
alvarosabu: @andretchen0 I just answered my own question, changed the `MeshPhongMaterial` on the playground for `Toon Material` and all good. Also with the default `MeshBasicMaterial` on the constructor

andretchen0: Hey @alvarosabu 

> I just answered my own question

Great! Is this issue resolved?

andretchen0: After the follow-up on Discord, I've added a default material.

alvarosabu (approved):

PR #570, opened 2024-12-20
fix: types

userquin:
This PR in draft since some vue dts files not being generated, once we generate them the types should be fine:

image


scripts/postbuild.mjs
```diff
@@ -0,0 +1,34 @@
+import { readFile, writeFile } from 'node:fs/promises'
+import { glob } from 'tinyglobby'
+
+/**
+ * Adds `.js` when importing the ./src/index.js in the dts
+ *
+ * @return {Promise<void>}
+ */
+async function patchRootDts() {
```
userquin, line 9: eliminar esto si no es necesario

scripts/postbuild.mjs
```diff
@@ -0,0 +1,34 @@
...
+        // console.log(file)
+        await writeFile(file.replace('.vue.d.ts', '.d.vue.ts'), await readFile(file, 'utf-8'), 'utf-8')
+    }
+}
+
+async function fixNode16() {
+    await patchVueDts()
+}
+
+fixNode16()
```
userquin, line 34: llamar directamente a `patchVueDts` y eliminar `fixNode16`

PR #574, opened 2024-12-25
fix(KeyboardControls): support non-qwerty keyboards (#573)

### LINKED ISSUE #573: Keyboard controls is improperly mapped
### Describe the bug

When you receive a key event, you receive two values, the event.code (KeyA) and the event.key (A).
The key is equal to the pressed letter, but letters change based on keyboard layout. 

While event.code don't.

Please swap the use of key to code, and use the proper values (A => KeyA).

Here is an example.
WhatsApp Image 2024-12-25 at 18 32 47

In my picture, WASD is as unusable as you can imagine, but if you use KeyW,  KeyA, KeyS, and KeyD, then it will map to my ZQSD no problem.

### Reproduction

https://cientos.tresjs.org/guide/controls/keyboard-controls.html

### Steps to reproduce

Access the webpage with a keyboard of different layout

### System Info

```shell
azerty layout keyboard
```

### Used Package Manager

npm

### Code of Conduct

andretchen0:
## Problem

#573

## Solution

Use key codes, e.g. `KeyW` rather than `w`. Closes #573 

## Problem

There are a few wording and formatting issues in the `<KeyboardControls />` docs page.

## Solution

Update docs.

### DIFF
--- a/docs/guide/controls/keyboard-controls.md
+++ b/docs/guide/controls/keyboard-controls.md
@@ -10,12 +10,10 @@
 * Arrow keys
 
 ::: info
-:memo: KeyboardControls uses `PointerLockControls` under the hood, meaning you can use [all the props from `<PointerLockControls />`](pointer-lock-controls#props) as well as it events.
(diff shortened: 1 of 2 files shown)


alvarosabu (approved):
@andretchen0 thanks for providing a fix and improving the docs

PR #579, opened 2025-01-04
feat: update dependencies and fix StatsGl component

alvarosabu:
- Upgraded `stats-gl` from version 2.0.1 to 3.6.0
- Added a temporary TypeScript comment in `StatsGl.ts` to handle an expected error while waiting for an update from `stats-gl`.

### DIFF
--- a/src/core/misc/StatsGl.ts
+++ b/src/core/misc/StatsGl.ts
@@ -93,6 +93,7 @@ export const StatsGl = defineComponent<StatsGlProps>({
     expose({ instance: statsGl })
 
     const node = document.body
+    // @ts-expect-error waiting for stats-gl update https://github.com/RenaudRohlinger/stats-gl/issues/16
     const statContainer = statsGl.dom || statsGl.container
 
     node?.appendChild(statContainer)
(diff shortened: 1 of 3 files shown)


src/core/misc/StatsGl.ts
```diff
@@ -93,6 +93,7 @@ export const StatsGl = defineComponent<StatsGlProps>({
     expose({ instance: statsGl })
 
     const node = document.body
+    // @ts-expect-error waiting for stats-gl update https://github.com/RenaudRohlinger/stats-gl/issues/16
     const statContainer = statsGl.dom || statsGl.container
```
andretchen0, line 97: tldr: This is a problem on the Cientos side. Not with `stats-gl`. Running `pnpm build` appears to clear the type error.

----

I can confirm that I also saw the type error if I removed the `@ts-expect-error` here. 

This was odd, because checking the installed exported types from `stats-gl` `3.6`, there's an public `dom` field.

Digging into it, after doing this initially ...

* Check out this PR
* `pnpm i`
* Right-click on `statsGl` below

... the editor showed a `2.4` version of the `Stats` type. That version doesn't have a public `dom` field. 

After running `pnpm build`, the `2.4` version was fully removed for me. Then the following line works in the editor without a `@ts-expect-error`

```ts
const statContainer = statsGl.dom
```

andretchen0 (changes requested):

PR #585, opened 2025-01-08
fix: add import to lensflare from three addons

alvarosabu:


andretchen0:
I created this issue on three-stdlib:

https://github.com/pmndrs/three-stdlib/issues/396

In the meantime, how about sticking `Lensflare.js` from /examples/jsm in Cientos until three-stdlib is fixed? That avoids build errors and works as expected.

CodyJasonBennett:
I fixed the upstream issue but would suggest importing from `three/addons` if possible since this is an ESM-only package. `three-stdlib` does tree shake completely, where `three/addons` does not. You can mitigate that by emitting separate JS files with `preserveModules` or similar and `package.sideEffects = false`, and bundlers won't include files unless referenced in code. Importing from `three/examples/jsm/**` creates hazards since some bundlers require explicit file extensions, which can be lost during transpilation, as it is technically source code.

PR #433, opened 2024-05-29
fix: 432 async draco loader

emavitta:
This fix #432, wrapping `new DRACOLoader()` in an async function to prevent racing condition, e.g. while looping through an array of models with draco compression.

PR #598, opened 2025-03-05
fix: 432 async draco loader

AntoninRousset:
This is a rebase of #433 which also makes the createDRACOLoader a sync function. The important point, I believe, is to not share `dracoLoader` accross `useGLTF` as it can get disposed while `useLoader` needs it.


alvarosabu:
This should be fixed by #610

PR #648, opened 2025-09-25
fix: missing importing watch for precipitation component

JaimeTorrealba:
Small PR to correct the missing import watch for precipitation component


alvarosabu (approved):
LGTM @JaimeTorrealba thanks

PR #650, opened 2025-09-26
fix: with the v5 the reflector component stop working.

JaimeTorrealba:
Problem: the way the color prop was being set

Solution: set the color prop manually using watch

https://github.com/user-attachments/assets/774f49fe-86cc-4e44-9569-798a34f33203


alvarosabu:
Mmm, I understand the fix, but I'm curious why it stopped working via the declarative prop `:material-uniforms-color-value`. Is this an underlying issue? 🤔

